PR #17, opened 2017-12-21
Implement Core Package

### LINKED ISSUE #3: Experiment Files
- Load the experiment file associated with the current experiment
- Experiment number can be specified in `cfg.yaml`
- Do not specify a file structure

Example entry:
```
experiment:
  - mfx12345
  - mfx25423
```

Example experiment file:
```
from ophyd import Device
from cool_module import do_scan

class SpecialClass(Device):
    def cool_scan():
        return do_scan()

special_object = SpecialClass()
__all__ = ['special_object']
```

### LINKED ISSUE #5: Hutch File
- Load from `hutch.py` in the same directory as `cfg.yaml`, if specified in the config.
- Do not enforce any particular structure

Example entry:
```
hutch:
    - hutch.py
```

### LINKED ISSUE #8: Namespaces
Provide a method to specify namespace groupings of devices in the `cfg.yaml`. This should be flexible enough to allow us to group objects by class, beamline location, and source file. This does not need to encompass small groups, but it should be easy to specify these in a `hutch.py` file.

### LINKED ISSUE #10: Solve Imports in User Files
User should have an easy way to pick objects from earlier load steps without accidentally creating an import loop.

### LINKED ISSUE #13: Installed Launch Script
At `setup.py` we need to specify a launch script that only takes a path to a config file. This will run the show.

ZLLentz:
## Description

I've written the core of this package, to the point where future PRs can address one issue at a time.
- Launchable script (`hutch-python`)
- Config files
- Load hutch.py
- Load experiment, given a name
- Organize namespaces by object type
- Refer to objects from previous stages of the load
- Travis CI unit tests
- Working conda recipe
- Interactive `tstpython`
- Cleaned-up logging

closes #3 
closes #5 
closes #8 
closes #10 
closes #13 

The docs still need a lot of work, but I'm opening an issue for it. Don't be intimidated by "3259 additions", most of that is just versioneer.

I have dropped Python 3.5 support because it appears as though my yaml dict keys in 3.5 were not guaranteed to be ordered correctly.


hutch_python/plugins/experiment.py
```diff
@@ -0,0 +1,82 @@
...
+from importlib import import_module
+from types import SimpleNamespace
+
+from ..base_plugin import BasePlugin
+from .. import utils
+
+logger = logging.getLogger(__name__)
+
+
+class Plugin(BasePlugin):
```
teddyrendahl: I'm not sure I like having multiple `Plugin` classes. Is there an issue with `ExperimentPlugin` ?

ZLLentz: It seemed verbose to me to refer to `experiment.ExperimentPlugin` rather than `experiment.Plugin`. It also made it annoying elsewhere to `getattr(module, plugin_name.Title() + 'Plugin')` instead of just `module.Plugin`

hutch_python/tests/conf.yaml
```diff
@@ -0,0 +1,48 @@
...
+  lcls: 1
+  platform: 4
+
+happi:
+  beamline:
+    - FEE
+    - HXD
+    - MFX
+
+lightpath: MFX
```
teddyrendahl: Lets let lightpath figure out the rest of the beamline? Seems redundant to have to the beamlines, hutch and lightpath specified

hutch_python/tests/conf.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+load:
+  - hutch.py
+
+experiment:
+  #name: automatic
+  name: mfxx000
+  import: experiment.User() as x
```
teddyrendahl: 👍

hutch_python/tests/conf.yaml
ZLLentz: I agree that this particular `conf.yaml` is very redundant. Is it still reasonable to allow us to specify other `happi` imports in the configuration file? I may end up wrapping the `lightpath` Plugin into the `hutch` Plugin.

hutch_python/plugins/experiment.py
teddyrendahl: Seems fair

hutch_python/tests/conf.yaml
teddyrendahl: That makes sense. I think standard deployment should be able to load everything just from hutch name. But you are correct, fancier things should be allowed.

PR #25, opened 2017-12-22
CI: Conda build and misc fixes

ZLLentz:
## Description

- Check the conda recipe as part of the CI
- Install the built package to pick up dependencies
- conda config --add channels
- Simplify the log initialization so that it works on more systems. This was causing my weird docs build error.

## Motivation and Context

- Avoid breaking conda recipe, encourage keeping the requirements accurate
- The travis yml file gets verbose with channels and run requirements, this eliminates that
- I like conda more than pip

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,15 @@ install:
   - conda update -q conda conda-build
   # Useful for debugging any issues with conda
   - conda info -a
-  # Grab all dependencies
-  # Build dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION --file dev-requirements.txt
-  - source activate test-environment
   # Test conda build
   - conda build -q conda-recipe
+  # Is it there?
+  - ls /home/travis/miniconda/conda-bld/linux-64
   # Install it
   - conda install -q hutch-python --use-local
+  # Create test environment
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION --file dev-requirements.txt --use-local
+  - source activate test-environment
 
 script:
   - coverage run run_tests.py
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -18,19 +18,23 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
+  - conda config --add channels conda-forge
+  - conda config --add channels skywalker-tag
+  - conda config --add channels lightsource2-tag
+  - conda install conda-build
+  - conda update -q conda conda-build
```
teddyrendahl: This is a nicer way to do this. Order added is the order the opposite order they are searched?

.travis.yml
```diff
@@ -18,19 +18,23 @@ install:
...
-  - conda install coverage pip wheel pytest flake8 codecov -c defaults -c conda-forge
-  # Module test dependencies
-  - conda install ipython pyyaml coloredlogs ophyd pcds-devices -c defaults -c conda-forge -c lightsource2-tag -c skywalker-tag
-  # Build it!
-  - python setup.py install
+  - conda install -q coverage pip wheel pytest flake8 codecov
+  # Test conda build
+  - conda build -q conda-recipe
+  # Install it
+  - conda install -q hutch-python --use-local
```
teddyrendahl: So this installs using the conda-recipe specs? I like this... Rather have an up to date conda-recipe than `requirements.txt`

.travis.yml
ZLLentz: I'll have to double-check whether they add to the front or the back of the queue.

Issue #27, opened 2018-01-16
Allow Plugins to request additional information


teddyrendahl:
## Expected Behavior

There may be use cases where there is information that is needed by multiple Plugins i.e experiment number. In the current schema, this info would have to be duplicated in the YAML file.

## Possible Solution

Add an extra class attribute to the `BasePlugin` called `extraneous`. This would point to other information in the YAML file that may not be underneath the Plugin explicitly. Then in `load_conf.py` the `extraneous` list is checked any additional information needed is added to the `info` dictionary directly.
 For example

### YAML file
```YAML
experiment_no : LJ001

happi:
         ...
```

### Happi Plugin
```python
class Plugin
     name = 'happi'
     extraneous = ['experiment_no']
```

ZLLentz:
An even simpler implementation might be to not require an `extraneous` argument at all, and let all plugins reference the original yaml dictionary. Solid idea.

PR #28, opened 2018-01-16
Happi plugin

### LINKED ISSUE #1: Happi Loading
- Interface with `happi` to load devices
- Read from the `cfg.yaml` file to determine which beamlines, stands, etc. to load
- Start with simple key matching

Example entry:
```
happi:
  beamline:
    - MFX
    - HXD
    - FEE
  name:
    - CXI PulsePicker
```

teddyrendahl:
## Description

Basic `happi` Plugin. 

The YAML spec looks like this:
```YAML
happi:
    filename: path/to/my_db.json
    requirements:
        active: True
        beamline: MFX
```
The plugin then; instantiates the `happi.Client`, finds the devices that match the requirements, and loads the appropriate Python class.

### Details
* Currently only works with JSON files. We can figure out a way to specify all the relevant MongoDB info when the time comes
* You can **not** enter a list of items for a single one to match i.e get devices from the MFX or HXD beamlines with -> beamlines = ['MFX', 'HXD']. We have a few choices here, allow certain fields to be lists and make the `Plugin` smarter or enhance the search capabilities in `happi`. Currently the `Client.search` will only match attributes that match the full list.
* We will probably create a second client for the questionnaire backend when that is ready. We will then need the experiment number from the other section of the YAML file. Currently that is not passed in `info`. I posed a possible fix here #27
* Not sure what to put for `Plugin.priority`

## Motivation and Context

Closes #1 

## How Has This Been Tested?

Loads properly from the `happi_db.json` file now included with the tests.

Added to the  `yaml_files.rst` section of the Sphinx docs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
   - conda install conda-build
   - conda update -q conda conda-build
   - conda config --append channels conda-forge
-  - conda config --append channels skywalker-tag
+  - conda config --append channels skywalker-dev
   - conda config --append channels lightsource2-tag
   # Useful for debugging any issues with conda
   - conda info -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,12 @@ env:
 matrix:
   include:
     - python: 3.6
-      env: BUILD_DOCS=1
+      env: 
+         - PCDS_CHANNEL=skywalker-tag
+    - python: 3.6
+      env:
+         - PCDS_CHANNEL=skywalker-dev
+         - BUILD_DOCS=1
 
 install:
   - sudo apt-get update
@@ -21,7 +26,7 @@ install:
   - conda install conda-build
   - conda update -q conda conda-build
   - conda config --append channels conda-forge
-  - conda config --append channels skywalker-dev
+  - conda config --append channels $PCDS_CHANNEL
   - conda config --append channels lightsource2-tag
   # Useful for debugging any issues with conda
   - conda info -a
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,6 @@ strictly required. Other than Python 3.6+, there are few requirements for
 running this module. These are small utilities:
 
 - ``ipython``, for improved interactive sessions
-- ``happi``, for device loading
 - ``pyyaml``, for reading config files
 - ``coloredlogs``, for colored logging
 
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,8 @@ At LCLS, there is little point to running this module without:
 If present, some modules will enable extra features, which will be listed
 below:
 
-- None, at the moment
+- `happi <https://github.com/slaclab/happi>`_ will enable device loading from a
+  happi database, and from the experiment questionaire.
 
 In the future, I plan to support the following extra features:
 
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,7 @@ strictly required. Other than Python 3.6+, there are few requirements for
 running this module. These are small utilities:
 
 - ``ipython``, for improved interactive sessions
+- ``happi``, for device loading
 - ``pyyaml``, for reading config files
 - ``coloredlogs``, for colored logging
 
@@ -40,7 +41,6 @@ below:
 
 In the future, I plan to support the following extra features:
 
-- `happi <https://github.com/slaclab/happi>`_ will enable device loading from a happi database, and from the experiment questionaire.
 - `lightpath <https://github.com/slaclab/lightpath>`_ will enable specification of a path object to include (requires happi).
 - ``pyfiglet`` will enable hutch banners (think big ``xpppython`` on startup)
 
--- a/hutch_python/plugins/happi.py
+++ b/hutch_python/plugins/happi.py
@@ -1,0 +1,31 @@
+import logging
+
+import happi
+from happi.loader import load_devices
+
+from ..base_plugin import BasePlugin
+
+logger = logging.getLogger(__name__)
+
+
+class Plugin(BasePlugin):
+    """
+    Plugin to load search information from happi
+    """
+    priority = 5
+    name = 'happi'
+
+    def get_objects(self):
+        # Load the happi Client. Use the configured kwargs to find the
+        # containers requested, and then instantiate the devices needed
+        _file = self.info['filename']
+        client = happi.Client(path=_file)
+        # If we have no requirements, we want all the devices. Otherwise use
+        # the listed requirements dictionary
+        reqs = self.info.get('requirements', {})
+        if not reqs:
+            containers = client.all_devices
+        else:
+            containers = client.search(**reqs)
+        dev_namespace = load_devices(*containers, pprint=True)
+        return dev_namespace.__dict__
--- a/hutch_python/plugins/happi.py
+++ b/hutch_python/plugins/happi.py
@@ -32,7 +32,6 @@ class Plugin(BasePlugin):
     """
     Plugin to load search information from happi
     """
-    priority = 5
     name = 'happi'
 
     def get_objects(self):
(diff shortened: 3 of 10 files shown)


README.rst
```diff
@@ -24,6 +24,7 @@ strictly required. Other than Python 3.6+, there are few requirements for
 running this module. These are small utilities:
 
 - ``ipython``, for improved interactive sessions
+- ``happi``, for device loading
```
ZLLentz: Is `happi` an explicit requirement now? If not, it belongs in a different list.

hutch_python/plugins/happi.py
```diff
@@ -0,0 +1,51 @@
...
+from ..base_plugin import BasePlugin
+
+logger = logging.getLogger(__name__)
+
+
+class Plugin(BasePlugin):
+    """
+    Plugin to load search information from happi
+    """
+    priority = 5
```
ZLLentz: No need to assign a priority unless it's really important that your plugin runs first/last or at a specific point in time

requirements.txt
```diff
@@ -1,3 +1,4 @@
 ipython
 pyyaml
 coloredlogs
+git+https://github.com/slaclab/happi#egg=happi
```
ZLLentz: Same as before, is this an explicit requirement? i.e. can I load the module in some capacity if I don't have `happi` installed? If not, it doesn't belong here.

hutch_python/plugins/happi.py
```diff
@@ -0,0 +1,51 @@
...
+
+Example
+-------
+.. code:: YAML
+
+    happi:
+        filename: path/to/my_db.json
+        requirements:
+            active: True
+            beamline: MFX
```
ZLLentz: Are we only supporting one source file and one filter?

.travis.yml
```diff
@@ -21,7 +21,7 @@ install:
   - conda install conda-build
   - conda update -q conda conda-build
   - conda config --append channels conda-forge
-  - conda config --append channels skywalker-tag
+  - conda config --append channels skywalker-dev
```
ZLLentz: I'd rather we add a `happi` release than rely on the dev `happi` channel

requirements.txt
teddyrendahl: You are correct. I'll change this.

.travis.yml
teddyrendahl: I kind of disagree. Seems like getting the full feature set for this plugin might take a few iterations in each library. I realize version numbers are cheap, but I don't think making a large number of pedantic versions is the way forward.

I'll yield for now.

hutch_python/plugins/happi.py
teddyrendahl: I think the plan for February should be one JSON file and the Questionnaire. You can have as many filters as you like, but only one set. You can also not have a list of acceptable values for the reasons mentioned above.

ZLLentz: I see now. I feel like the plugin has limited use then, kind of just a "Get me my hutch stuff" button. That's still useful.

hutch_python/plugins/happi.py
ZLLentz: Actually, to expand a bit, assigning a priority when it isn't needed can cause confusion.

.travis.yml
ZLLentz: I see. I guess I made the assumption that `happi` was at a release point if we were ready to pull it into what will become a versioned `hutch-python` by the end of the month.

hutch_python/plugins/happi.py
ZLLentz: Still developing thoughts on this... The appropriate activate timing for this plugin is probably "before `lightpath`". So the priority should come before the default priority and after the namespace priority, so I think 5 is good.

I'm considering ditching this system and moving to a more fluid cause-and-effect system where you can specify which plugins need to be run before this one can be run.

teddyrendahl:
@ZLLentz This is failing because we specify `pcds-devices` in the conda-recipe. I think when we moved that repository it doesn't update our `skywalker-*` conda channels. Because we import an older pcds-devices it requires an older version of happi.  Why do we need the pcds-devices depndency there?

.travis.yml
teddyrendahl: We are close. Ideally happi `1.0.0` will work with pcds-devices `1.0.0`. Unfortunately, I need to wait to fix the containers until we fix the devices ...  It also seems to make sense to have the Questionnaire in there for 1.0.0 but this I'm less picky about

hutch_python/plugins/happi.py
teddyrendahl: I'll come back for this when I do the `lightpath` plugin

hutch_python/plugins/happi.py
teddyrendahl: Yeah unfortunately we are limited by the `happi` search. What is annoying is that MongoDb has a bunch of nice search features, but if we want `happi` to be backend agnostic it has to be able to everything it does for Mongo but with a JSON backend. That means writing the search code ourselves. It has been on the to-do list but I always wanted to wait for a need instead of a want. We can discuss if we need more complication than this. I really intended this as just a first pass M.V.P effort.

ZLLentz:
@teddyrendahl think of the conda recipe as "modules that we will always run with at lcls". It's not a dependency for this module but it's a dependency for running `mfxpython`. I made this choice because I wasn't intending to make conda recipes for the individual hutch ecosystems, and as a side-effect it makes it easy to test both the conda recipe and the full test suite with all plugins available. It also catches errors that are only visible from the top-level with all modules present, such as accidentally messing up the CI when we moved the module over.

I'm going to rule out making conda recipes for each hutch, but we could take the other approach and make a separate hutch python metapackage that requires this module, `pcds-devices`, and some others for this purpose. This isn't something I would have thought of before our nsls-ii trip. Do you think that's a good idea? If we had that, the metapackage build would be dead (indicating the issue you saw) but this module would let us proceed without being annoying.

ZLLentz:
@teddyrendahl Are you planning to tag a happi release or no?

ZLLentz:
@teddyrendahl If no happi release in the next 2 days, then this is good to go, but I'll open an issue to switch back to the skywalker-tag channel (or to a brand new channel?) before `1.0.0`

teddyrendahl:
Alright sorry for spamming. The conda-recipe specified `pcds-devices` but the package is now called `pcdsdevices`. Therefore it was finding the last build where the dash was in the package name which specified that happi had to be `0.4.0`.  I updated the tests to use the new `pcdsdevices` imports and changed the package name in the recipe. In general I would like to avoid pulling in this package for the tests if it isn't necessary. I realize that it looks good for the `test.yaml` to be a good example for the real thing, but it makes an unnecessary dependency for this test suite.

I'm on the fence for the `happi` release. Using `-tag` is more ideal but I'm not sure it is realistic while we break API for 2/3 of our packages in the month of January. 🤷‍♂️

ZLLentz:
I think you are 100% correct about not including pcdsdevices in the tests. I'll add an issue about pulling it out of this package entirely and start brainstorming names for the metapackage.

README.rst
```diff
@@ -40,7 +40,6 @@ below:
 
 In the future, I plan to support the following extra features:
 
-- `happi <https://github.com/slaclab/happi>`_ will enable device loading from a happi database, and from the experiment questionaire.
```
ZLLentz: Actually, one more nitpick: this line was originally moved to the top section, and now fully deleted.  It should have been moved into the middle section.

teddyrendahl: 👍

teddyrendahl:
Alright. Ready for merge provided we are fine with `-tag` designation for now.

teddyrendahl:
Deceptively. This should be ready for merge now. Instead of deciding whether to test on `-tag` or `-dev` I think the answer should be both! This way we recognize that this version of hutch python will need 1.0.0 at the minimum and will not work with older builds of happi. However, we have verified that with the newer happi versions the code in the Pull Request works.

ZLLentz:
This is all good. The tag builds are still broken, but we'll move forward ignoring that until we have things tagged down properly.

PR #31, opened 2018-01-26
ENH: More Info for Plugins

### LINKED ISSUE #27: Allow Plugins to request additional information

### LINKED ISSUE #29: Remove pcdsdevices from tests and recipe
## Expected Behavior

Tests should minimize their requirements.

## Current Behavior

Tests needlessly import `pcdsdevices`. This is because `pcdsdevices` is included for a more macro-level package management reasons.

## Possible Solution

Switch tests to group devices from arbitrary modules... literally any built-in. Come up with some other module to do the macro-level package management.

ZLLentz:
## Description

Plugins are now passed the entire yaml dict. They store this under `self.conf`, and break out their own piece of the dict as `self.info` for convenience.

All bonus dependencies have been removed from the tests. `happi` is still required in the recipe that ships with the module because there is a `happi` plugin, and I expect `happi` to play a role in the upcoming `hutch` plugin.

For the sake of testing, `PCDS_CHANNEL` was swapped from `skywalker-` to `pcds-`

## Motivation and Context

closes #27 
closes #29 

## How Has This Been Tested?

100% coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ matrix:
   include:
     - python: 3.6
       env: 
-         - PCDS_CHANNEL=skywalker-tag
+         - PCDS_CHANNEL=pcds-tag
     - python: 3.6
       env:
-         - PCDS_CHANNEL=skywalker-dev
+         - PCDS_CHANNEL=pcds-dev
          - BUILD_DOCS=1
 
 install:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -20,8 +20,7 @@ requirements:
     - ipython
     - pyyaml
     - coloredlogs
-    - happi >=0.5.0
-    - pcdsdevices >=0.3.0
+    - happi >=0.5.0  # Required if using happi plugin
 
 test:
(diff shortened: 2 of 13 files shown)


teddyrendahl:
Kicked Travis again with `happi` released as `v1.0.0rc1` and everything passes. Much appreciate the quick fix. I think as a design principle we'll be better off if we try and keep the information in `conf` much larger than the information in `info`, but that is mainly for the template discussion.

PR #37, opened 2018-01-30
ENH: Allow Metaplugins

ZLLentz:
## Description

Overriding `BasePlugin.pre_plugins` to return a list of plugins executes those plugins first, in the order provided.

## Motivation and Context

This allows us to build arbitrary compound plugins from smaller plugin building blocks.

## How Has This Been Tested?

No new tests, not covered. This will be covered for free when we write a metaplugin, and the implementation is simple enough that I'm not worried.

### DIFF
--- a/hutch_python/base_plugin.py
+++ b/hutch_python/base_plugin.py
@@ -18,15 +18,22 @@ class BasePlugin:
     priority = 0
     name = None
 
-    def __init__(self, conf):
+    def __init__(self, conf=None, info=None):
         """
+        Arguments will be passed every time by the loader, but are optional for
+        ease of testing and to help create metaplugins. At least one or the
+        other must be provided.
+
         Parameters
         ----------
-        conf: dict
+        conf: dict, optional
             The full dict from the yaml file.
+
+        info: object, optional
+            The portion of the dict specific to this plugin
         """
         self.conf = conf
-        self.info = conf[self.name]
+        self.info = info or conf[self.name]
 
     def get_objects(self):
         """
@@ -38,6 +45,17 @@ class BasePlugin:
             Mapping from final global reference name to object
         """
         pass
+
+    def other_plugins(self):
+        """
+        Return a list of instantiated plugins that must be run before running
+        this plugin.
+
+        Returns
+        -------
+        plugins: list
+        """
+        return []
 
     def future_object_hook(self, name, obj):
         """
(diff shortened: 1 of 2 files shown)


hutch_python/base_plugin.py
```diff
@@ -18,15 +18,22 @@ class BasePlugin:
...
-        conf: dict
+        conf: dict, optional
             The full dict from the yaml file.
+
+        info: object, optional
+            The portion of the dict specific to this plugin
         """
         self.conf = conf
-        self.info = conf[self.name]
+        self.info = info or conf[self.name]
```
teddyrendahl: I think this should be
```python
info or conf.get(self.name, {})
```
The current setup would raise a KeyError if `info` is not passed and `Plugin.name` is not in `conf`. That means for instance you couldn't pass this directly to Questionnaire Plugin
```
experiment:
   run : 15
   proposal: X123
```

hutch_python/load_conf.py
```diff
@@ -62,16 +62,18 @@ def get_plugins(conf):
...
     plugins = defaultdict(list)
 
-    for plugin_name in conf.keys():
+    for plugin_name, info in conf.items():
         try:
             module = import_module('hutch_python.plugins.' + plugin_name)
         except ImportError:
             module = None
             err = 'Plugin {} is not available, skipping'
             logger.warning(err.format(plugin_name))
```
teddyrendahl: On an unrelated note this should probably be bumped to error.

teddyrendahl:
Not bad. A few nitpicks but this looks good.

hutch_python/base_plugin.py
ZLLentz: You're right that this is a problem. I'll change it to `self.info = info or conf.get(self.name)` tomorrow.  The default `None` return is fine here because there is no guarantee as to the type of `info`, it can be `dict`, `list`, or `str` depending on the plugin.

hutch_python/load_conf.py
ZLLentz: We should talk about what we want each of the log levels to mean, since they pop up in different colors during the load. Maybe something like:
```
Error (red): Malformed configurations, user-caused exceptions, etc.
Warning (yellow): Device failed to load
```
Eh, maybe that's not quite right either. The point I'm trying to make is that it's not obvious to me which of these elevated log statements should be red vs yellow.

hutch_python/base_plugin.py
teddyrendahl: 👍

PR #38, opened 2018-01-30
ENH: Beamline Plugin

ZLLentz:
## Description

Add plugin that loads the `Daq` and `RE` objects, and calls the `happi` plugin to load beamline devices.

requires #37

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,27 +19,20 @@ Requirements
 ------------
 
 This module is intended to be run using the latest pcds release in the
-`pcds-envs <https://github.com/pcdshub/pcds-envs>`_ package, but this is not
-strictly required. Other than Python 3.6+, there are few requirements for
-running this module. These are small utilities:
+`pcds-envs <https://github.com/pcdshub/pcds-envs>`_ package.
+
+This module requires Python 3.6+ and the following utilities:
 
 - ``ipython``, for improved interactive sessions
 - ``pyyaml``, for reading config files
 - ``coloredlogs``, for colored logging
-
-At LCLS, there is little point to running this module without:
-
 - `pcdsdevices <https://github.com/pcdshub/pcdsdevices>`_ for our Device abstraction layers
 - ``pydaq`` for running the DAQ
+- `happi <https://github.com/slaclab/happi>`_ will enable device loading from a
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -19,27 +19,20 @@ Requirements
...
-If present, some modules will enable extra features, which will be listed
-below:
-
 - `happi <https://github.com/slaclab/happi>`_ will enable device loading from a
-  happi database, and from the experiment questionaire.
+   happi database, and from the experiment questionaire.
+- `bluesky <https://github.com/nsls-ii/bluesky>`_ for scanning
 
-In the future, I plan to support the following extra features:
+In the future, we plan to support the following extra features:
```
teddyrendahl: 😸

hutch_python/plugins/beamline.py
```diff
@@ -0,0 +1,32 @@
...
+        # Make the Daq object
+        hutch = self.info.lower()
+        if SIM_DAQ:
+            Cls = SimDaq
+        else:
+            Cls = Daq
+        daq = Cls(name='daq', platform=DAQ_MAP[hutch])
+        # Make the Daq RunEngine
+        RE = make_daq_run_engine(daq)
+        return dict(daq=daq, RE=RE)
```
teddyrendahl: How does this work in the current hutch python? Does every session connect to pydaq? Can you connect from machines where the DAQ is not running?

teddyrendahl:
Kicked `pcdsdevices` on Travis again, now passing on `-dev` and `-tag`

hutch_python/plugins/beamline.py
ZLLentz: You can import `pydaq` without connecting to the DAQ, but you have to avoid using it. I was thinking we'd work toward either a command-line option to swap to `SimDaq` instead of `Daq` for these other machines (helpful for debugging), or we can try to dynamically decide which machines should load the real daq... Actually, this isn't quite good enough, since what we *really* want is a way to have this return just a vanilla RE for non-daq work.

ZLLentz:
I'll work on resolving the merge conflicts tomorrow, it's probably something simple.

PR #41, opened 2018-01-30
ENH: Experimental Plugin as MetaPlugin

teddyrendahl:
## Description

Load information from the Questionnaire as a pre-plugin step to the Experiment.  This can be disabled by setting `questionnaire: False` in the experiment section.

## How Has This Been Tested?

Added a new test to `test_plugins/test_experiment.py`

Added documentation on how to disable in the Experimental Plugin


hutch_python/plugins/experiment.py
```diff
@@ -11,9 +11,18 @@
...
     """
     name = 'experiment'
 
+    def pre_plugins(self):
+        if self.info.get('questionnaire', True):
+            from .questionnaire import Plugin as QSPlugin
+            return [QSPlugin(conf=self.conf)]
+
     def get_objects(self):
         expname = self.info['name']
```
ZLLentz: Can you change this to be looking for `run` and `proposal` and combining them into the `expname`? Otherwise this doesn't actually work with the config we were looking to use for mfx.

teddyrendahl:
Rebased to solve merge conflict with the `mfx0000.py` file.

PR #43, opened 2018-01-30
ENH: Namespace Rework and Metadata Namespace

### LINKED ISSUE #40: Metaplugins Break Top-level Namespacing
If two versions of a plugin get loaded, the second overwrites the first in the `hutch_python.__init__` loaded namespaces.

ZLLentz:
Implement the metadata-based namespace config from the happi metadata. In the process, invert how the namespace module works. I'm not super happy with how this turned out, because it's no longer possible to import these namespaces into user code, but it should be ok for tomorrow.

closes #40


hutch_python/tests/test_plugins/test_questionnaire.py
```diff
@@ -42,3 +43,12 @@ def test_questionnaire_plugin():
...
     assert objs['inj_x'].proposal == 'LR12'
+
+
+def test_questionnaire_bad_conf():
+    logger.debug('test_questionnaire_bad_conf')
+    conf = {'experiment': {}, 'questionnaire': True}
+    plugin = Plugin(conf)
+    hutch_python.plugins.questionnaire.QSBackend = QSBackend
+    with pytest.raises(ValueError):
+        plugin.get_objects()
```
teddyrendahl: What is this here for? Just 100% coverage?

ZLLentz: If it raises a different error, there is a problem...

teddyrendahl:
Seems like it works. We may want to simplify this in the future and just hard-code this if we don't find ourselves changing it often. Otherwise, seems fine.

I think there are a few things we can do to make this a little cleaner and reduce the work you have to do in this plugin along.

1) The lightpath should tell you what beamlines you expect
2.) Happi should tell you what stands you expect.
3.) I also want to think about this looks as just a function that scrapes a collection of devices. Not as they're loaded, but from the IPython namespace for instance. This would mean that we could make the collections during an IPython session even.

Good enough for initial rollout. Interested to hear scientist feedback before we figure out how much to flush out this feature or not.

PR #45, opened 2018-01-31
FIX: Build fixes and misc naming

ZLLentz:
## Description

- Change `.travis.yml` to ensure that it runs tests using the installed version of the package
- move `logging.yml` into the distribution and make sure it is included in the build
- settle on a convention for a `beamline.py` file and a `hutch` plugin

## Motivation and Context

- If the tests don't pass on the conda-build version, the module is broken. This will help us catch things in the future (and it helped me debug this too)
- the module needs a logging config to run, so we include it
- people are used to `beamline.py` and it was foolish of me to try to change that.


ZLLentz:
HMM, maybe I revert this to get coverage back up

teddyrendahl:
Not in love with testing twice but 🤷‍♂️

PR #46, opened 2018-01-31
WIP: More Fixes

ZLLentz:
## Description

Work in progress, accumulating changes as we find issues live.

- catch errors in pre_plugin load
- allow the use of a default class checking path for the namespace plugins
- add HUTCH BANNERS


ZLLentz:
@teddyrendahl added hutch banners to this PR, need to go back later to restore coverage.

PR #48, opened 2018-02-02
Authentication Configuration

teddyrendahl:
## Description

The `opr` accounts need to have a username and password to login to the LCLS Questionnaires. There isn't a super elegant solution to this problem; we don't want to hard-code the password places, we don't want operators to have to login to hutch python.

This solution searches for a `qs.cfg` file, either in the current directory, hidden in the current directory, or hidden in the user directory. If this file is found, the information contained is used to authenticate. If not found, the user's kerberos ticket is used. Not an incredible solution but it means that:

1) Regular users will just use their Kerberos without thinking about it as long as they don't have a `qs.cfg` file in their home directory
2) Operators can access the Questionnaire and their password is safely hidden in a read-only file

## How Has This Been Tested?

Added tests for both authentication methods

Added documentation on what the configuration file should look like.

### DIFF
--- a/hutch_python/tests/conftest.py
+++ b/hutch_python/tests/conftest.py
@@ -5,6 +5,8 @@ from copy import copy
 from contextlib import contextmanager
 from collections import namedtuple
 
+import pytest
+
 # We need to have the tests directory importable to match what we'd have in a
 # real hutch-python install
 sys.path.insert(0, os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


hutch_python/tests/conftest.py
```diff
@@ -33,29 +35,55 @@ def restore_logging():
...
+            'args': ['{{run}}', '{{proposal}}',
+                     '{{user}}', '{{pw}}', '{{kerberos}}'],
             'kwargs': {},
             'name': 'inj_x',
             'prefix': 'TST:USR:MMN:01',
             'purpose': 'Injector X',
             'type': 'Device',
             'run': self.run,
+            'user': self.user,
+            'pw': self.pw,
```
ZLLentz: It's moderately alarming if these are stored somewhere in the device info.

teddyrendahl: This is all for testing **only**. The issue is that the `QSBackend` is a local variable in `Plugin.get_objects()`. It is therefore hard to introspect and see that it was setup properly. To get around this I made a fake object that stores the password it was instantiated that way I can just check the output of `get_objects` for the state of the client. In practice, this information is dropped after `get_objects` as they are only local variables.

ZLLentz: Sounds good, just looked suspect when I was reading down the line.

PR #49, opened 2018-02-03
LOG: Let's log... almost everything

ZLLentz:
## Description

- log all IPython input and output in the debug stream
- log all exceptions that bubble up to the top of the IPython terminal to the debug stream
- remove file handlers for all log levels except for debug
- one log file per session
- fix codecov that I broke

## Motivation and Context

This will be very useful for tracking down some of the more obscure bugs when we're trying to figure out how to reproduce them. The redundant logfiles were redundant. One file per session means no user vs user conflicts, and it means we can log two terminal sessions by the same user independently.

## How Has This Been Tested?

`tstpython` passes all the spot-tests, and the automated tests pass.

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -2,6 +2,10 @@ import os
 import sys
 import argparse
 
+from IPython.terminal.embed import (InteractiveShellEmbed,
+                                    load_default_config)
+
+from .ipython_log import init_ipython_logger
 from .load_conf import load
 from .log_setup import setup_logging
 from .plugins import hutch
@@ -28,3 +32,19 @@ def setup_cli_env():
 
     # Load objects from the configuration file
     return load(args.cfg)
+
+
+def hutch_ipython_embed():
+    """
+    This is very hacky, but I couldn't find a better way to adjust the shell
+    from a call to embed.
+    """
+    config = load_default_config()
+    config.InteractiveShellEmbed = config.TerminalInteractiveShell
+    frame = sys._getframe(1)
+    shell = InteractiveShellEmbed.instance(_init_location_id='%s:%s' % (
+        frame.f_code.co_filename, frame.f_lineno), config=config)
+    init_ipython_logger(shell)
+    shell(header=u'', stack_depth=2, compile_flags=None,
+          call_location_id='%s:%s' % (frame.f_code.co_filename,
+                                      frame.f_lineno))
--- a/hutch_python/ipython_log.py
+++ b/hutch_python/ipython_log.py
@@ -1,0 +1,43 @@
+"""
+IPython plugin to log inputs, outputs, and tracebacks to the debug stream.
+"""
+import sys
+import traceback
+import logging
+
+import IPython
+
+import __main__
+
+logger = logging.getLogger(__name__)
+
+
+class IPythonLogger:
+    def __init__(self):
+        self.prev_err = None
+
+    def log(self):
+        line_num = len(__main__.In) - 1
+        if line_num == 0:
+            return
+        last_in = __main__.In[-1]
+        logger.debug('In  [{}]: {}'.format(line_num, last_in))
+        try:
+            last_out = __main__.Out[line_num]
+            # Convert to string, limit to max tweet length
+            last_out = str(last_out)[:280]
+            logger.debug('Out [{}]: {}'.format(line_num, last_out))
+        except KeyError:
+            pass
+        global PREV_ERR
+        if hasattr(sys, 'last_value') and sys.last_value != self.prev_err:
+            tb = ''.join(traceback.format_exception(sys.last_type,
+                                                    sys.last_value,
+                                                    sys.last_traceback))
+            logger.debug('Exception in IPython session, traceback:\n' + tb)
+            self.prev_err = sys.last_value
+
+
+def init_ipython_logger():
+    ip = IPython.get_ipython()
+    ip.events.register('post_execute', IPythonLogger().log)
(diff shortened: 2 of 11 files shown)


hutch_python/ipython_log.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def log(self):
+        line_num = len(__main__.In) - 1
+        if line_num == 0:
+            return
+        last_in = __main__.In[-1]
+        logger.debug('In  [{}]: {}'.format(line_num, last_in))
+        try:
+            last_out = __main__.Out[line_num]
+            # Convert to string, limit to max tweet length
```
teddyrendahl: 😆

hutch_python/ipython_log.py
```diff
@@ -0,0 +1,43 @@
...
+            tb = ''.join(traceback.format_exception(sys.last_type,
+                                                    sys.last_value,
+                                                    sys.last_traceback))
+            logger.debug('Exception in IPython session, traceback:\n' + tb)
+            self.prev_err = sys.last_value
+
+
+def init_ipython_logger():
+    ip = IPython.get_ipython()
+    ip.events.register('post_execute', IPythonLogger().log)
```
teddyrendahl: 👏 This is cool!

teddyrendahl:
Looks cool. You managed to start an IPython shell in Travis 🥇 

I really like the idea. I assume you looked at the built-in IPython logging at decided to go this direction. I also support the idea of user based logs. The multiple log files for level seemed redundant.

https://ipython.org/ipython-doc/3/interactive/reference.html#session-logging-and-restoring

Any reason we aren't just going this route?

ZLLentz:
Full disclosure: I didn't dive into the IPython docs at all before writing this. I just wanted to have some fun. I don't think those session logs have timestamps, Out statements, or the tracebacks, but I'll double-check before going further.

I'm glad I can break Travis every so often.

ZLLentz:
Reminder to self: do something different when the error is `RuntimeError: maximum recusion depth exceeded`

ZLLentz:
@teddyrendahl It looks like In, Out, and timestamps are all built-in to the `ipython` file logger, if you give it the right options. I'll have to consider the merits of switching to that for the In/Out and leaving this for the Exception logging. This built-in `ipython` logger is mainly intended so that you can re-run your session (it outputs a `.py` file)

Pros:
- uses built-in features
- probably less buggy
- less issues getting the launcher to work correctly

Cons:
- separate files for in/out and everything else

I also want to explore if it's possible to pipe the outputs of the built-in logger to `logger.debug` to remove the one con.

ZLLentz:
@teddyrendahl I think this is ready now

hutch_python/ipython_log.py
```diff
@@ -0,0 +1,42 @@
...
+                # Convert to string, limit to max tweet length
+                last_out = str(last_out)[:280]
+                logger.debug('Out [{}]: {}'.format(line_num, last_out))
+            except KeyError:
+                pass
+            if hasattr(sys, 'last_value') and sys.last_value != self.prev_err:
+                tb = ''.join(traceback.format_exception(sys.last_type,
+                                                        sys.last_value,
+                                                        sys.last_traceback))
+                logger.debug('Exception in IPython session, traceback:\n' + tb)
```
teddyrendahl: Keeping these as `debug` even though we have an exception in the terminal?

hutch_python/cli.py
```diff
@@ -28,3 +32,19 @@ def setup_cli_env():
...
+    """
+    config = load_default_config()
+    config.InteractiveShellEmbed = config.TerminalInteractiveShell
+    frame = sys._getframe(1)
+    shell = InteractiveShellEmbed.instance(_init_location_id='%s:%s' % (
+        frame.f_code.co_filename, frame.f_lineno), config=config)
+    init_ipython_logger(shell)
+    shell(header=u'', stack_depth=2, compile_flags=None,
+          call_location_id='%s:%s' % (frame.f_code.co_filename,
+                                      frame.f_lineno))
```
teddyrendahl: This scares me a little bit. Largely because I have no idea what it is doing. Let's leave it in, but I might be quick to blame it if funky stuff starts happening in our IPython windows.

teddyrendahl:
I think this looks good enough. As stated above, I want to be careful we don't end up messing up our IPython shell, but until we actually see evidence of that, I believe it is worth a shot as this will be incredibly useful for diagnosing bugs.

As far as an implementation detail, I'm tempted to make a unique logging level for the input commands. That way they are very easy to pull out of the log. It seems like that'll be the common workflow for this. It wouldn't be hard to imagine have a custom bug report script that does things like _bug report with last 10 user inputs_. Obviously you can check for the `IN` tag, but the custom level seems nice. Just a thought.

hutch_python/cli.py
ZLLentz: copy pasted from ipython and then removed all the unneeded lines

hutch_python/ipython_log.py
ZLLentz: logger is configured to print info and above, so logger.error or warning here gives us a double print of the exception

ZLLentz:
Very good point about the custom log level. I'm not sure how to implement that, but I'd like to do it before the PR if it's easy. It'd be called INPUT and be above DEBUG and below INFO.

ZLLentz:
https://docs.python.org/3/library/logging.html#logging.addLevelName

hutch_python/ipython_log.py
teddyrendahl: That makes sense. Too crazy to add another logging level huh ...?

ZLLentz: It wouldn't be crazy to add yet another logging level, so we have INPUT and INPUTERROR.

PR #50, opened 2018-02-07
ENH: Logging Features

ZLLentz:
## Description

Adds a few debug features to user space, the most important being:
- availability of debug mode, where the console logger is at level `DEBUG`
- command-line option `--debug` to start in debug mode
- `_debug_mode()` returns `True` or `False`, can be set as `_debug_mode(True)`
- `_debug_context()` to temporarily enable debug mode for a block of code

Changes:
- change `INPUT` to level 5 (less then `DEBUG`)
- group log files in directories by month
- make sure log directories have full write permissions

## Motivation and Context

- can easily enable `DEBUG` prints in a session to see what is happening without repeatedly referring to a log file
- `INPUT` being below `DEBUG` allows the previous feature to be used without an input/output echo
- log files in one flat directory will be a disaster by the end of the year

## How Has This Been Tested?

Covered in automated tests

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -7,7 +7,7 @@ from IPython.terminal.embed import (InteractiveShellEmbed,
 
 from .ipython_log import init_ipython_logger
 from .load_conf import load
-from .log_setup import setup_logging
+from .log_setup import setup_logging, set_console_level
 from .plugins import hutch
 
 
@@ -31,7 +31,12 @@ def setup_cli_env():
     hutch.HAPPI_DB = args.db
 
     # Load objects from the configuration file
-    return load(args.cfg)
+    objs = load(args.cfg)
+
+    # Add cli debug tools
+    objs['_debug_console_level'] = set_console_level
+
+    return objs
 
 
 def hutch_ipython_embed():
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -7,7 +7,8 @@ from IPython.terminal.embed import (InteractiveShellEmbed,
 
 from .ipython_log import init_ipython_logger
 from .load_conf import load
-from .log_setup import setup_logging, set_console_level
+from .log_setup import (setup_logging, set_console_level, debug_mode,
+                        debug_context, debug_wrapper)
 from .plugins import hutch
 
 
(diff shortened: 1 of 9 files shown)


hutch_python/cli.py
```diff
@@ -31,7 +31,12 @@ def setup_cli_env():
     hutch.HAPPI_DB = args.db
 
     # Load objects from the configuration file
-    return load(args.cfg)
+    objs = load(args.cfg)
+
+    # Add cli debug tools
+    objs['_debug_console_level'] = set_console_level
+
```
teddyrendahl: Any thought about putting a `--debug` mode arg. That way you can set up debug logging in the terminal even during device loading? 

```python
if args.debug:
     set_console_level(logging.DEBUG)
objs = load(args.cfg)
```
I know how you feel about command line args ...

teddyrendahl:
This is awesome. One question I'm pondering is how accessible do we need to make these functions? I'm tempted to have a `debug_mode` function that handles setting the logging level in one line. Right now a scientist would still need to import `logging`, import the `hutch_python` helper function then call it. This seems like a lot to expect from someone not intimate with Python details. Maybe just a `debug_mode` wrapper is valuable. 

I think the argument against this, is that only experts will want to print the `DEBUG` info to the screen. A bunch of log messages might only serve to confuse people.

hutch_python/cli.py
ZLLentz: That's a good idea, will implement

ZLLentz:
It's a little easier than that, with this implementation a `_debug_console_level` function is always available in the namespace, no imports required. I figured it would be nice to include these debug helper functions as `_debug_{something}` so they are all grouped together and you won't stumble upon them while looking for normal objects. You are right if you mean that having something that's just `_debug_mode(debug=True)` is more useful than remembering the numbers 10 and 20. You're also right if you mean that having a `_debug_wrapper` or `_debug_context` would be useful, so you can do things like:
```
_debug_wrapper(my_function, *args, **kwargs)

with _debug_context:
    my_function()
    another_function()
```
How much of this should we include?

ZLLentz:
I ended up including all of it because it was straightforward. Updated PR description.

teddyrendahl:
Not sure if underscore is necessary. Time will tell.

PR #51, opened 2018-02-08
ENH: Easy reference to database objects

ZLLentz:
## Description

Make all loaded objects available from `hutchname.db`, e.g. we can now do `from mfx.db import mfx_attenuator`.

The `hutchname/db.py` file will be autogenerated and annotated.

## Motivation and Context

It needs to be very very easy to track what objects are being loaded from the databases and to access them.

## How Has This Been Tested?

The tst experiment file gets objects from `tst.db`. It was also tested in a test ipython session.

Docstrings updated.

`db.py`
```
"""
The objects referenced in this file are populated by the tstpython
initialization. If you wish to use devices from this file, import
them from tst.db after calling the tstpython startup script.

hutch-python last loaded on 2018-02-08 14:16:21.154315
with the following objects:

tst_device_1         <class 'happi.device.Device'>
tst_device_2         <class 'happi.device.Device'>
daq                  <class 'pcdsdevices.daq.Daq'>
RE                   <class 'bluesky.run_engine.RunEngine'>
UniqueDevice         <class 'type'>
calc_thing           <class 'function'>
unique_device        <class 'beamline.UniqueDevice'>
x                    <class 'experiments.mfxx010.User'>
fn                   <class 'types.SimpleNamespace'>
funcs                <class 'types.SimpleNamespace'>
tst                  <class 'types.SimpleNamespace'>
"""
```

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -67,6 +67,8 @@ def read_conf(conf, filename=None):
         conf_path = Path(filename)
         hutch_path = conf_path.parent
         objects = run_plugins(plugins, hutch=hutch, hutch_path=hutch_path)
+    else:
+        logger.warning('hutch was not defined in yaml!')
     return objects
 
 
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -173,7 +173,8 @@ def run_plugins(plugins, hutch=None, hutch_path=None):
     if do_db:
         quotes = '"""\n'
         header = 'Automatically generated file, do not edit.\n\n'
-        body = 'hutch-python last loaded on {}\nwith the following objects:\n\n'
+        body = ('hutch-python last loaded on {}\n'
+                'with the following objects:\n\n')
         text = quotes + header + body.format(datetime.datetime.now())
         for name, obj in all_objs.items():
             text += '{:<20} {}\n'.format(name, obj.__class__)
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -172,7 +172,10 @@ def run_plugins(plugins, hutch=None, hutch_path=None):
     # Annotate db file at the end
     if do_db:
         quotes = '"""\n'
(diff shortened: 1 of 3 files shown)


hutch_python/load_conf.py
```diff
@@ -103,13 +115,31 @@ def run_plugins(plugins):
...
     objs: dict{str: object}
         Return value of load
     """
     all_objs = {}
 
+    # Dummy module for easier user imports
+    do_db = None not in (hutch, hutch_path)
+    if do_db:
+        db_module_name = hutch + '.db'
+        db_path = hutch_path / hutch / 'db.py'
```
teddyrendahl: Whoah ... `pathlib` is neat

teddyrendahl:
This looks very cool. 

My one request is kind of nitpicky. Can you expand on the string explanation inside the file? If someone looks in there I would expect that they are very lost and need guidance. So something along the lines of:

```python
"""
The objects referenced in this file are populated by the `hutch-python` initialization.
If you wish to use devices from this file, import them from mfx.db after calling 
the hutch-python startup script.
"""

File: hutch_python/load_conf.py [hunk 2]
```diff
@@ -37,24 +39,34 @@ def load(filename):
...
     hutch_python.clear_load()
     plugins = get_plugins(conf)
-    objects = run_plugins(plugins)
+    hutch = conf.get('hutch')
+    if filename is None:
+        objects = run_plugins(plugins)
+    elif hutch is not None:
+        conf_path = Path(filename)
+        hutch_path = conf_path.parent
+        objects = run_plugins(plugins, hutch=hutch, hutch_path=hutch_path)
```
teddyrendahl: Might be nice to just throw a warning here on else:
```python
"Neither a configuration file or hutch was specified. Unable to load plugins"
```
Just scares me that if `hutch` is not entered properly this does nothing and does so silently

ZLLentz:
@teddyrendahl I did those two changes

PR #52, opened 2018-02-08
ENH: Enable hutch-python scripts

ZLLentz:
## Description

Adding any script e.g. `script.py` as a positional argument to `hutch-python` will run the script after loading the environment instead of dropping us into an `ipython` terminal.

## Motivation and Context

Some scripts need objects defined and loaded by `hutch-python`, and there wasn't a convenient way to run these sorts of scripts. This doesn't cover the situation where you want to run in the same environment as your hutch python but without the objects, which will be handled in the template for each hutch.

## How Has This Been Tested?

Added tests to `test_cli.py` for coverage and added `test_tstpython.py` for a script-level sanity check.

It hasn't been documented yet. Sue me.


teddyrendahl:
Looks good.

At some point we might want to add a way to run a script then enter the IPython terminal. Until we have a use case, lets ignore it.

PR #53, opened 2018-02-09
ENH: Add and use iterable namespaces

ZLLentz:
## Description

Use `IterableNamespace` instead of `SimpleNamespace`.

## Motivation and Context

This allows us to use statements such as `for motor in motors:` and `list(slits)`. This was low priority but it was implementable within 10mins so I just went for it.

### DIFF
--- a/hutch_python/plugins/experiment.py
+++ b/hutch_python/plugins/experiment.py
@@ -1,9 +1,8 @@
 import logging
 from importlib import import_module
-from types import SimpleNamespace
 
 from ..base_plugin import BasePlugin
-from .. import utils
+from ..utils import extract_objs, IterableNamespace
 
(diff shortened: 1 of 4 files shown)


teddyrendahl:
I wonder if we should make sure that it is iterated in a specific order? I think as written it will just be in the order things were added to the namespace which is kind of obtuse. Can we alphabetically by key?

ZLLentz:
You're right, as written it will be the order we added them. It'd be easy to do something like:
```
objs = [v for v in self.__dict__.values()]
yield from sorted(objs, key=lambda x: x.name) 
```

ZLLentz:
I settled on the less verbose:
```
for _, obj in sorted(self.__dict__.items()):
    yield obj
```

PR #55, opened 2018-02-09
TST: Fix test so it works with happi that caches devices

ZLLentz:
## Description

Clear the cache before each test.

## Motivation and Context

`happi` caches devices in `pcds-dev` now, so the later tests that use the same device name are picking up the old devices in our dev build.


ZLLentz:
@teddyrendahl I'm merging this on your day off to fix the `pcds-dev` build. Take a look next week and enjoy your weekend.

PR #58, opened 2018-02-27
MNT: Massive Refactor

### LINKED ISSUE #4: Auto Select Experiment File
- Select current experiment by specifying `automatic` in `cfg.yaml`
- Requires some work porting things to python 3. May be easy, may be hard, has not been evaluated.

Example entry:
```
experiment:
  - automatic
```

### LINKED ISSUE #56: Fix/Change how we do db.py
- `chmod` a+w so multi users are guaranteed to be able to rewrite it
- switch to `db.txt` and a virtual `hutch.db` module to remove possibility of corruption

### LINKED ISSUE #57: Do not report error without experiment file
## Current Behavior

Currently if there is no experiment file an error is printed at startup.
```
  INFO     Loading experiment ls0816
  ERROR    Plugin experiment failed to load, skipping
```
## Requested Behavior
Many experiments (at least thus far in MFX) do not need experiment files. They are content with just the devices within the `happi` database and the Questionnaire. This error message appears to freak scientists out. I think we should downgrade the not-exists error message, and only report a full exception if there is a file that exists that can not be imported properly.

ZLLentz:
## Description

- Generate `db.txt` instead of `db.py` to prevent user from messing self up. Load from a virtual `hutch.db` module instead.
- Remove the top-level `hutch_python` cache in favor of a `LoadCache` class for simpler implementation
- Remove the concept of a plugin in favor of greater load process control
- Include `happi` db path inside of `conf.yaml` instead of as a separate command-line argument
- Allow loads without `cfg` arg as a side effect of better bad argument handling. This opens a mostly empty session.
- Collect all constants in one `constants.py` file because I don't want to keep remembering which files define which constants
- Split up what were previously plugins into small files that get the relevant objects. These are imported and called in `load_conf`
- Move scattered utils into `utils.py`. This may need to be split up into a `utils` directory in the future.
- Stop supporting many different experiment file structures: it's easier for us and for the users to have a framework where you create a `User` class and no other options or ambiguity
- Simplify log handling slightly and allow a no-file mode for hutchless sessions
- Present `namespace` construction macros that users can use
- Include three namespaces by default: `p`/`plans`, `m`/`motors`, `s`/`slits`, `a`/`all_objects`
- Remove `conf.yaml` namespace config

## Motivation and Context

- Things are hard to implement in `BasePlugin` framework
- Keep the load process consistent across hutches
- Remove user's ability to mess themselves up with the `conf.yaml` file
closes #57 
closes #56 
closes #4 

## How Has This Been Tested?

- Tests were overhauled, but still some work to be done here
- Requires the daq updates from https://github.com/pcdshub/pcdsdevices/pull/155


hutch_python/utils.py
```diff
@@ -1,8 +1,55 @@
...
+from .constants import (CUR_EXP_SCRIPT, CLASS_SEARCH_PATH, HUTCH_COLORS,
+                        SUCCESS_LEVEL)
+
+logging.addLevelName('SUCCESS', SUCCESS_LEVEL)
 logger = logging.getLogger(__name__)
+logger.success = partial(logger.log, SUCCESS_LEVEL)
+
+
+@contextmanager
+def safe_load(name, cls=None):
```
teddyrendahl: 👍

hutch_python/namespace.py
```diff
@@ -0,0 +1,108 @@
...
+        elif isinstance(obj, cls):
+            include = True
+        if include:
+            setattr(class_space, name, obj)
+            logger.debug('Include %s in cls=%s namespace', name, cls)
+
+    return class_space
+
+
+def metadata_namespace(md, scope=None):
```
teddyrendahl: This is so much clearer. We can even have scientists write these live or in the `beamline.py` file

hutch_python/load_conf.py
```diff
@@ -1,199 +1,228 @@
...
+        else:
+            hutch_banner(banner.lower())
+        conf_path = Path(cfg)
+        hutch_dir = conf_path.parent
+        return load_conf(conf, hutch_dir=hutch_dir)
+
+
+def load_conf(conf, hutch_dir=None):
     """
+    Step through the objcet loading procedure, consulting conf as needed.
```
teddyrendahl: typo "objcet"

hutch_python/exp_load.py
```diff
@@ -0,0 +1,28 @@
+import logging
+from importlib import import_module
+
+from .utils import safe_load
+
+logger = logging.getLogger(__name__)
+
+
+def get_exp_objs(proposal, run):
```
teddyrendahl: Enforcing a way to do this is a good idea. Standardization makes for easier maintainable code top to bottom

hutch_python/load_conf.py
```diff
@@ -157,6 +158,8 @@ def load_conf(conf, hutch_dir=None):
 
     # Make RunEngine
     RE = RunEngine({})
+    bec = BestEffortCallback()
+    RE.subscribe(bec)
```
teddyrendahl: Should we also install the `qt_kicker` here?

ZLLentz: Good catch

ZLLentz:
@teddyrendahl it turns out that `install_qt_kicker` can't run on `Travis`. I will wrap in a try/except block, even though it's only for the CI...

hutch_python/load_conf.py
teddyrendahl: If you want it to pass:
```
before_script:
    #Take from docs.travis-ci.com/user/gui-and-headless-browsers
  - "export DISPLAY=:99.0"
  - "sh -e /etc/init.d/xvfb start"
  - sleep 3 # give xvfb some time to start
```
Otherwise the try/except works

ZLLentz:
The dev build now passes, the tag build still fails as expected.

PR #60, opened 2018-03-01
ENH: Lightpath

teddyrendahl:
## Description

Loads a `lightpath.BeamPath` object with the name `{hutch}_beampath`. Adds a dependency on the `lightpath` module which in turn needs `PyDM`

## How Has This Been Tested?

* Added a specific test for the `get_lightpath` function
* Added a check in `test_load_conf` that it gets executed properly when loading the entire configuration

Updated module docstring for `hutch_python.happi`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
   - conda config --append channels conda-forge
   - conda config --append channels $PCDS_CHANNEL
   - conda config --append channels pcds-tag
+  - conda config --append channels pydm-tag
   - conda config --append channels lightsource2-tag
   # Useful for debugging any issues with conda
   - conda info -a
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -28,6 +28,7 @@ install:
   - conda config --append channels conda-forge
   - conda config --append channels $PCDS_CHANNEL
   - conda config --append channels pcds-tag
+  - conda config --append channels pydm-tag
```
ZLLentz: Stuff like this makes me want to move sooner on filling our channel

teddyrendahl: Yup. Noted in https://github.com/pcdshub/typhon/pull/45 as well

PR #61, opened 2018-03-05
ENH: Device status shower

ZLLentz:
This is more of a proposal/idea than a PR. It came out of a discussion I had with @apra93 today about issues with `__repr__` in split and delay.

I think we should avoid changing device `__repr__` to large status readbacks because it has repercussions in tons of miscellaneous python code.

Instead, here is a potential option where we give our devices `status` methods that an `ipython` event hook notices and prints. After a line like `mfx_att` has been executed, `ipython` calls anything we've registered to the `post_execute` hook. At this point, the last element of the `Out` dictionary is the object we just referenced, `mfx_att`. This gives us an opportunity to print the status.

There might be better ways to achieve the goal of just being able to type the object and getting information about its status, but I thought this was a reasonably good enough place to start. Let me know what you think, we can let this sit for a while. I'll add tests if we think this is a good idea.

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -7,6 +7,7 @@ from IPython.terminal.embed import InteractiveShellEmbed
 
 from .daq import set_daq_sim
 from .ipython_log import init_ipython_logger
+from .ipython_status import init_ipython_status
 from .load_conf import load
 from .log_setup import (setup_logging, set_console_level, debug_mode,
                         debug_context, debug_wrapper)
@@ -80,6 +81,7 @@ def hutch_ipython_embed(stack_offset=0):
     # + stack_offset for extra levels between this call and user space
     shell = InteractiveShellEmbed.instance()
     init_ipython_logger(shell)
+    init_ipython_status(shell)
     shell(stack_depth=stack_depth)
 
 
--- a/hutch_python/ipython_status.py
+++ b/hutch_python/ipython_status.py
@@ -1,0 +1,28 @@
+"""
+IPython plugin to print device.status() after showing the repr as Out in
+interactive sessions
+"""
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class IPythonStatus:
+    def __init__(self, ipython):
+        self.In = ipython.user_ns['In']
+        self.Out = ipython.user_ns['Out']
+
+    def show_status(self):
+        index = len(self.In) - 1
+        if index in self.Out:
+            obj = self.Out[index]
+            if hasattr(obj, 'status'):
+                try:
+                    print(obj.status())
+                except Exception:
+                    err = 'Error showing device status'
+                    logger.debug(err, exc_info=True)
+
+
+def init_ipython_status(ip):
+    ip.events.register('post_execute', IPythonStatus(ip).show_status)


teddyrendahl:
I'm not super against this. I'm going to offer a counter solution for discussion. In [pcdsdevices](https://github.com/pcdshub/pcdsdevices/blob/c5b1586bff14ff027b1f29570dd14ac72843abe6/pcdsdevices/mv_interface.py#L44) we have a `MvInterface`. You could imagine just a regular `PCDSInterface` where calling your device gives you the summary print out.  Right now on `MvInterface` we have an empty call set to to return the position of the motor, but I think it would be reasonable to extend this to be a full print out. This solution seems cleaner in the sense that we are not doing `IPython` shenanigans but it does require subclassing all our devices...

As far as the print out itself, `ophyd.Device` has a built in `summary`. It looks like this for `ophyd.EpicsMotor`
```
data keys (* hints)
-------------------
*my_motor
 my_motor_user_setpoint

read attrs
----------
user_readback        EpicsSignalRO       ('my_motor')
user_setpoint        EpicsSignal         ('my_motor_user_setpoint')

config keys
-----------
my_motor_acceleration
my_motor_motor_egu
my_motor_user_offset
my_motor_user_offset_dir
my_motor_velocity

configuration attrs
----------
motor_egu            EpicsSignal         ('my_motor_motor_egu')
velocity             EpicsSignal         ('my_motor_velocity')
acceleration         EpicsSignal         ('my_motor_acceleration')
user_offset          EpicsSignal         ('my_motor_user_offset')
user_offset_dir      EpicsSignal         ('my_motor_user_offset_dir')

Unused attrs
------------
offset_freeze_switch EpicsSignal         ('my_motor_offset_freeze_switch')
set_use_switch       EpicsSignal         ('my_motor_set_use_switch')
motor_is_moving      EpicsSignalRO       ('my_motor_motor_is_moving')
motor_done_move      EpicsSignalRO       ('my_motor_motor_done_move')
high_limit_switch    EpicsSignal         ('my_motor_high_limit_switch')
low_limit_switch     EpicsSignal         ('my_motor_low_limit_switch')
direction_of_travel  EpicsSignal         ('my_motor_direction_of_travel')
motor_stop           EpicsSignal         ('my_motor_motor_stop')
home_forward         EpicsSignal         ('my_motor_home_forward')
home_reverse         EpicsSignal         ('my_motor_home_reverse')


```
It'd be nice if this gave the current values of all these signals, maybe as as an additional option to `summary`. I want to make sure this section makes it back in to the core `ophyd` library.

ZLLentz:
I think it's unreasonable to assume that no devices will want to use their `__call__` methods for things other than status printouts, but I agree that this is cleaner than relying on `ipython` shell mods for something that should be simple. Not sure yet.

I agree that generating these statuses should usually be something programmatic, ideally a built-in, but for some things we're going to want custom setups (e.g. the little ascii art in the old hutch python for the attenuators)

apra93:
I like the `IPython` solution only because it's really simple to implement, but that it doesn't travel with the device makes it an incomplete solution. At the same time, using the `__call__` method doesn't seem like a good enough solution because if we will need to retrain them to make the object call, we may as well just go all the way and use a method specific to this functionality like `status()` or `st()`.

Also, with `summary`, is your idea that we change the printout to be the that from `status`?

teddyrendahl:
Yeah. I don't think it would be crazy to add a `summary(live=True)` to output something that looks very similar to `status`. Our classes that have very special statuses (like the attenuators) could just override this.

I think it fundamentally should live on the device not in IPython. What if I want to print a summary to the Elog or something?

ZLLentz:
`elog.post(device_name.live_summary())`
`print(device_name.live_summary())`

The `IPython` mod is simply to enable the behavior where you can type:
`device_name`

And that's it, to get the summary.

PR #63, opened 2018-03-06
ENH: Built-in Cookie-Cutter Template

ZLLentz:
## Description

`hutch-python --create hutch` makes a `hutch` directory.

## Motivation and Context

This should be easy. It now is.

## How Has This Been Tested?

Added a test and spent an extra two hours finding the only way to make Python accept these jinja'd filenames in the manifest.in asdfasefbarhdf


teddyrendahl:
Nice! I imagined this as a separate script `hutch-create`. Any reason why you just put it as a keyword?

ZLLentz:
No particular reason. It was easier to add an argument to an existing script than to create a new script and set up the parsing again. It wasn't clear which was better so I picked fewer lines of code.

ZLLentz:
It's not hard to invert this if you think it's better to keep the reader script `hutch-python` separate from miscellaneous utilities. I'd suggest a more general `hutch-python-utils --create hutchname` or something to that effect rather than create a new script every time we want a helper feature. It's not obvious to me whether piling options into one script or splitting them into two is better.

ZLLentz:
Last two changes were a hard-coded db path and a basic `set-dev` script for checking out repos and putting them in a `dev` folder such that they will override the python imports.

ZLLentz:
The fix for the dev travis is in #66

ZLLentz:
A last minute thought: are the names `xxxpython`, `xxxrun`, etc. optimal? Perhaps each script in each hutch repo should have the same launch scripts with the same names to make them easier to manage via an automated update using github.

PR #71, opened 2018-03-10
DOC: Doc. Docs.

ZLLentz:
## Description

It's docs. I think this is a good first pass, everything is filled out except for the `Tutorial` page.

Minor changes:
- The `daq` setup was double-caching the `RE`. It wasn't causing any issues, but I removed it so I could give a shorter accurate return description in the docs.
- Make the argument parser outside of the cli setup function so it could be included in the docstring.
- Fix a test that broke randomly. It looked like a `Python` bug. Did it in a simpler way this time.

## Motivation and Context

Docs.

## How Has This Been Tested?

The docs.

It is docs.


README.rst
```diff
@@ -21,22 +20,33 @@ This module is intended to be run using the latest pcds release in the
...
+- ``pyami`` for viewing data from the daq
+
+These are not available outside of the slac intranet, and in some cases
+must be sync'd with the active DAQ versions, independent of all other
+packages.
+
+To automatically select the current experiment, you must run with access to
+the LCLS NFS directories.
 
-This module is not yet ready for users. When completed, the included conda
```
teddyrendahl: 🍾

docs/source/conf.py
```diff
@@ -75,7 +75,7 @@
 exclude_patterns = ['_build']
 
 # The reST default role (used for this markup: `text`) to use for all documents.
-# default_role = None
+default_role = 'any'
```
teddyrendahl: What does this do? Just curious.

teddyrendahl:
👏 👏 👏 

Way to go above the call of duty. These look great.

docs/source/conf.py
ZLLentz: It defines the default link directive as `any`, so every time you do something like
```
`text`
```
In anything (docstrings, sphinx docs, etc.) it tries to find a valid link target somewhere in your documentation, whether it be a page, a class, a label, etc.

E.g. you don't have to put :directives: in your docstrings to link to functions

teddyrendahl: 😮

PR #84, opened 2018-03-15
ENH: Include questionnaire objects in the user object

### LINKED ISSUE #82: Add Questionnaire Motors to Experiment Object

ZLLentz:
## Description

- After loading ``User()``, add the questionnaire objects to it
- Load ``SimpleNamespace()`` as a fallback for a missing or malformed experiment file
- alias ``x`` to ``user``

## Motivation and Context

This is a feature of the old python that is very helpful. The ``x`` or ``user`` object becomes the object that the users are allowed to call methods from during their experiment.

closes #82 

## How Has This Been Tested?

I added a check to make sure that the experiment object contained a reference to a questionnaire object.

I've added to the docs as appropriate on the startup routine page and on the experiments page. I have built the docs in my environment and they look as intended.

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -228,8 +228,10 @@ def load_conf(conf, hutch_dir=None):
     if proposal is not None:
         qs_objs = get_qs_objs(proposal, run)
         cache(**qs_objs)
-        exp_objs = get_exp_objs(proposal, run)
-        cache(**exp_objs)
+        user = get_exp_objs(proposal, run)
+        for name, obj in qs_objs.items():
+            setattr(user, name, obj)
+        cache(x=user, user=user)
 
     # Default namespaces
     with safe_load('default groups'):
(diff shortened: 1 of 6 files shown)


hutch_python/load_conf.py
```diff
@@ -228,8 +228,10 @@ def load_conf(conf, hutch_dir=None):
     if proposal is not None:
         qs_objs = get_qs_objs(proposal, run)
         cache(**qs_objs)
-        exp_objs = get_exp_objs(proposal, run)
-        cache(**exp_objs)
+        user = get_exp_objs(proposal, run)
+        for name, obj in qs_objs.items():
+            setattr(user, name, obj)
```
teddyrendahl: Should we at all be concerned about us getting a `User` object we can't set attributes on? Or I'm just being paranoid.

ZLLentz: I didn't consider this. How would this happen?

teddyrendahl: If someone made object inherit from some strange object. Probably not worth worrying about.

ZLLentz: I suppose I can imagine something a little less malicious, like:

```
@property
def inj_x(self):
    return something
```

But I think for these instances we simply say... don't do that!

PR #91, opened 2018-03-21
MAINT: Rename qs.cfg to web.cfg

teddyrendahl:
## Description

Originally the configuration file was for the Questionnaire alone so it was given the specific name `qs.cfg`. In reality this information is useful for all the `pswww` applications so we should look for a more general name like `web.cfg`.
 
## How Has This Been Tested?

Tests are now done with `web.cfg` file

References to `qs.cfg` have been changed to `web.cfg`


teddyrendahl:
Agreed. We might need a better system for delaying docs to make sure that our online documentation stays in lock with our released environments in general. We'll see how often this comes up.

PR #94, opened 2018-03-29
FIX: Make db.txt file all-write upon creation

ZLLentz:


hutch_python/cache.py
```diff
@@ -67,6 +67,9 @@ def write_file(self):
                     + body.format(datetime.datetime.now()))
             for name, obj in self.objs.__dict__.items():
                 text += '{:<20} {}\n'.format(name, obj.__class__)
+            if not db_path.exists():
+                db_path.touch()
+                db_path.chmod(0o666)
```
teddyrendahl: Run the `chmod` outside the if statement?

ZLLentz: Fails if not already chmoded

ZLLentz: e.g. if you didn't have permission to write before, you don't have permission to chmod now

PR #95, opened 2018-03-29
ENH: Set up the presets

ZLLentz:
## Description

Use the presets recently added to `pcdsdevices`. Create the `presets/beamline` and `presets/proposalrunnum` directories.

## Motivation and Context

The people have spoken. We've split between the `beamline` presets that are permanent and the `experiment` presets that are not.

## How Has This Been Tested?

Assume presets work in `pcdsdevices`, make sure we've set up the paths here.

In `pcdsdevices`

This will definitely fail on travis until we cut a new release for `pcdsdevices` with the presets.

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -264,10 +264,10 @@ def load_conf(conf, hutch_dir=None):
                     path.mkdir()
                     path.chmod(0o777)
             if proposal is None:
-                setup_preset_paths(beamline=beamline_presets)
+                setup_preset_paths(hutch=beamline_presets)
             else:
-                setup_preset_paths(beamline=beamline_presets,
-                                   experiment=experiment_presets)
+                setup_preset_paths(hutch=beamline_presets,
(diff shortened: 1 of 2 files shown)


ZLLentz:
I'd like to have your opinions on:

- paths, are `presets/beamline` and `presets/ls2516` good for the folders?
- function calls, currently the set is:
```
motor.presets.add_beamline
motor.presets.add_experiment
motor.presets.add_here_beamline
motor.presets.add_here_experiment
```
But I'm not sure if these should be shorter or changed, etc. of they are ok as is.

teddyrendahl:
Tempted to change `add_experiment` to `add_exp`? Just a thought. Can't think of an appropriate shortening of beamline ... `add_hutch`. I don't know. 

I'm fine with the `YAML` file paths.

ZLLentz:
Your shortenings seem appropriate. Let's sit on this just a bit longer, I want to get it in right before cutting the pcds-envs release.

PR #96, opened 2018-03-30
ENH: Include Subcomponent in Class Namespace

ZLLentz:
## Description

When defining our scope for the purposes of the `class_namespace`, expand the scope to include subdevices.

## Motivation and Context

If we're searching for `EpicsMotor` objects we should consider the `x`, `y`, and `z` motors of a top-level multi-axis device even if we haven't defined them explicitly at the top level of our scope.

## How Has This Been Tested?

A test has been added.

Docstring has been extended.

Note that I was originally intending this to be a much more ambitious PR where we could check attributes of everything and spider out through the whole python namespace, but this caused... problems.

### DIFF
--- a/hutch_python/namespace.py
+++ b/hutch_python/namespace.py
@@ -4,6 +4,8 @@ This module provides utilities for grouping objects into namespaces.
 from inspect import isfunction
 import logging
 
+from ophyd import Device
+
 from .utils import (IterableNamespace, find_class, strip_prefix,
                     extract_objs)
 
@@ -22,6 +24,10 @@ def class_namespace(cls, scope=None):
         Every object attached to the given modules will be considered for the
         `class_namespace`. If ``scope`` is omitted, we'll check all objects
         loaded by ``hutch-python`` and everything in the caller's global frame.
+        If anything is an instance of ``ophyd.Device``, we'll also include the
+        object's components as part of the scope, using the ``name`` attribute
+        to identify them rather than the attribute name on the device. This
+        will continue recursively.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


hutch_python/namespace.py
```diff
@@ -41,6 +47,23 @@ def class_namespace(cls, scope=None):
...
+        if obj not in cache:
+            cache.add(obj)
+            for comp_name in getattr(obj, 'component_names', []):
+                sub_obj = getattr(obj, comp_name)
+                accumulate(sub_obj, scope_objs, cache)
+            # Don't accidentally override
+            if obj.name not in scope_objs:
+                scope_objs[obj.name] = obj
+
+    for name, obj in scope_objs.copy().items():
```
teddyrendahl: Why copy?

teddyrendahl:
I'm concerned that by calling `getattr` on all signals you are forcing connection / creation on all of our devices. This could severely affect load time.

hutch_python/namespace.py
ZLLentz: To avoid mutating a dictionary that we're iterating through

ZLLentz:
Since we're using `lazy=False` on everything, this doesn't force any connections (except on the upcoming lazy magic components). They've already instantiated and started their connections before this. The connections themselves do not add much to the load time because they are not blocking, it's the `__init__` statements and making the objects that add up.

I included the signals because there is a public attribute that tells me all the component names, while the attribute that contains only the devices is private.

The motivation behind this PR is that you want your namespace groups to be complete without having to remake any subdevices that you'd like to be included. Once we put, say, `xpp_sb2_pim_y` onto the `xpp_sb2_pim` object, we expect that our group of motors has the pim motor on it, otherwise grouping by class this way was somewhat pointless. The `ophyd` structure is nice, but it's not the only logical grouping.

PR #98, opened 2018-04-04
CI: Switch to our channel

ZLLentz:


teddyrendahl:
You still want to build this twice?

ZLLentz:
I think the rest of the CI is fine. The double test isn't kosher but it catches issues with the installer.

PR #99, opened 2018-04-05
ENH: Daq Platform rework

ZLLentz:
## Description

Concept for doing the daq platform in the yaml file instead of hard-coded in the module. This allows the user to configure the daq platform on a per-host or per-hutch basis.

## Motivation and Context

Cxi has 2 daqs

## How Has This Been Tested?

I will add tests before the merge

I have updated the docs


teddyrendahl:
Agreed. This will break MFX if we upgrade without adding the platform default of 4. Not a problem just a note.

PR #105, opened 2018-04-09
ENH: Elog

### LINKED ISSUE #100: Add basic ELog object
## Expected Behavior

Add an `elog` object to the global namespace. Possible now that we have `elog` v0.1.0`

## Current Behavior

No `elog` capabilities.

teddyrendahl:
## Description

Create an `elog.HutchELog` object named `elog` in the standard load setup.

### Nothing is Easy
I thought this was going to be a brainless PR until I realized that I need to determine which experiment the specific machine I am working on belongs to. i.e I do not want to accidentally post to the main experiment logbook when I am working on the secondary one.

The best way I thought of to accomplish this is to simply check if we are using the "default" `daq_platform`. We can then make the assumption that the default DAQ means primary experiment and any special platform instructions indicate that we should use the secondary experiment. This will break in the following regimes:
* The configuration inverts the intended YAML configuration and uses the secondary DAQ platform as the default.
* There are 3 DAQs in a hutch ...
* We are only running one experiment and a user logs on from a machine with special platform instructions.

Out of these the third one is the only one that actually concerns me. I'm just not sure how we get around this problem. If a machine is sometimes used as the secondary DAQ sometimes as just a support machine for the primary experiment and the `conf.yaml` is not edited, how will we ever know?
 
## Motivation and Context

Closes #100 
## How Has This Been Tested?

Tests added with mock ELog. 

**TBD**

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -10,8 +10,8 @@ from socket import gethostname
 from bluesky import RunEngine
 from bluesky.callbacks.best_effort import BestEffortCallback
 from bluesky.utils import install_kicker
-from pcdsdevices.mv_interface import setup_preset_paths
 from elog import HutchELog
+from pcdsdevices.mv_interface import setup_preset_paths
 
 from . import plan_defaults
 from .cache import LoadCache
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -174,12 +174,16 @@ def load_conf(conf, hutch_dir=None):
                          'file.'))
 
     try:
+        # This is an internal variable here for note-keeping. The ELog uses
+        # this to determine if we are in the secondary or primary DAQ mode
+        default_platform = True
         platform_info = conf['daq_platform']
         hostname = gethostname()
         try:
             daq_platform = platform_info[hostname]
             logger.info('Selected %s daq platform: %s',
                         hostname, daq_platform)
+            default_platform = False
         except KeyError:
             daq_platform = platform_info['default']
             logger.info('Selected default %s daq platform: %s',
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -234,7 +234,7 @@ def load_conf(conf, hutch_dir=None):
         else:
             logger.info("Configuring ELog to post to secondary experiment")
             kwargs = {'station': '1'}
-        cache(**{'elog': HutchELog.from_conf(hutch.upper(), **kwargs)})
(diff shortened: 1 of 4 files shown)


hutch_python/load_conf.py
```diff
@@ -219,6 +224,18 @@ def load_conf(conf, hutch_dir=None):
             bp = get_lightpath(db, hutch)
             cache(**{"{}_beampath".format(hutch.lower()): bp})
 
+    # Elog
+    with safe_load('elog'):
+        # Use the fact if we we used the default_platform or not to decide
+        # whether we are in a specialty station or not
+        if not default_platform:
```
ZLLentz: Dumb comment warning: the `if not ... else` pattern is my pet peeve because `if ... else` is clearer 99% of the time (just reversing the code blocks). You don't have to change this, I just wanted to mention it.

hutch_python/load_conf.py
```diff
@@ -219,6 +224,18 @@ def load_conf(conf, hutch_dir=None):
...
+            logger.info("Configuring ELog to post to secondary experiment")
+            kwargs = {'station': '1'}
+        else:
+            logger.debug("Using primary experiment ELog")
+            kwargs = dict()
+        cache(**{'elog': HutchELog.from_conf(hutch.upper(), **kwargs)})
```
ZLLentz: could have been `cache(elog=...)` instead of undoing your own `{}` with a `**`

hutch_python/tests/test_load_conf.py
```diff
@@ -57,6 +57,26 @@ def test_conf_platform():
...
+    assert objs['elog'].pw == 'pw'
+    # Define default platform
+    objs = load_conf({'daq_platform': {'default': 1},
+                      'hutch': 'TST'})
+    assert objs['elog'].station is None
+    # Define host platform
+    hostname = gethostname()
+    objs = load_conf({'daq_platform': {'default': 3,
+                                       hostname: 4},
+                      'hutch': 'TST'})
```
ZLLentz: :+1:

hutch_python/load_conf.py
```diff
@@ -11,6 +11,7 @@
 from bluesky.callbacks.best_effort import BestEffortCallback
 from bluesky.utils import install_kicker
 from pcdsdevices.mv_interface import setup_preset_paths
+from elog import HutchELog
```
ZLLentz: Ok, this is even more dumb than the other things, but can you put this into the correct alphabetical spot? It's going to bug me forever having one thing out of order.

PR #108, opened 2018-04-12
ENH: Load upstream devices

### LINKED ISSUE #69: Inclusion of preceding beamlines
## Expected Behavior

Now that `lightpath` has a hardcoded configuration we can tell all the devices that need to be loaded quite easily. It would be nice if downstream beamlines are included as well, maybe contained within a namespace i.e
```python
mfx_dg1_pim
xrt.hxx_valve
```
## Current Behavior

Only devices on the beamline that match the hutch are included.

teddyrendahl:
## Description

Use the `lightpath` config to check what the upstream beamlines are and the range we care about them. From this we can load **all** the upstream devices along the beamline not just those with the same name as the hutch.

#### Note
This does not work quite right with XCS. We need to tell it to load **both** the `PBT` and standard `XCS` paths. I'm not sure the best way to complete this task. Tempted to just load the few devices on the `PBT` from `happi` but in the `beamline.py` file ....

## Motivation and Context

Closes #69 

## How Has This Been Tested?

Added a test where we load only upstream devices. Tested this in `MFX` and everything worked exactly as it should.

### DIFF
--- a/hutch_python/happi.py
+++ b/hutch_python/happi.py
@@ -2,6 +2,7 @@ import logging
 
 import happi
 import lightpath
+from lightpath.config import beamlines
 from happi.loader import load_devices
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


ZLLentz:
I'm also not sure about XCS. I think your suggestion is a good start until we've thought it over more. Having two paths is annoying.

This implementation is clean. I'll pull it in tomorrow morning unless you have second thoughts.

PR #114, opened 2018-04-23
DOC: Preset docs

ZLLentz:
## Description

- Add docs for presets usage
- Simplify docs makefiles
- Add `elog` to docs requirements section
- Fix a typo


ZLLentz:
@teddyrendahl Take a look when you get back and see if I missed something important to include here

PR #119, opened 2018-04-26
API: Simplify metadata_namespace into tree_namespace

ZLLentz:
## Description

- Stop using happi md to build the tree
- Use only the names

## Motivation and Context

https://github.com/pcdshub/Bug-Reports/issues/2
I was trying to resolve this issue, and I found it to be extremely unwieldy to partially match metadata and names... And there was a lot of extra code for stripping the names but only if the name matched the metadata, etc... So why not simplify the whole process? This is much simpler code to follow and much easier for a user to take advantage of and manipulate.

## How Has This Been Tested?

Modified the tests and checked in mfx

Probably should be added explicitly in the docs, but the docstrings have been updated.

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -20,7 +20,7 @@ from .constants import VALID_KEYS
 from .daq import get_daq_objs
 from .exp_load import get_exp_objs
 from .happi import get_happi_objs, get_lightpath
-from .namespace import class_namespace, metadata_namespace
+from .namespace import class_namespace, tree_namespace
 from .qs_load import get_qs_objs
 from .user_load import get_user_objs
 from .utils import (get_current_experiment, safe_load, hutch_banner,
@@ -277,10 +277,9 @@ def load_conf(conf, hutch_dir=None):
         default_class_namespace('EpicsMotor', 'motors', cache)
         default_class_namespace('Slits', 'slits', cache)
         if hutch is not None:
-            meta = metadata_namespace(['beamline', 'stand'],
-                                      scope='hutch_python.db')
+            tree = tree_namespace(scope='hutch_python.db')
             # Prune meta, remove branches with only one object
-            for name, space in meta.__dict__.items():
+            for name, space in tree.__dict__.items():
                 if count_ns_leaves(space) > 1:
                     cache(**{name: space})
 
--- a/hutch_python/namespace.py
+++ b/hutch_python/namespace.py
@@ -111,59 +111,43 @@ def class_namespace(cls, scope=None):
     return class_space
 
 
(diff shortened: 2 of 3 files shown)


teddyrendahl:
At first read I'm 👎. Then I realized that you squashed a bug by deleting almost a hundred lines of code and now I am reconsidering. I see the benefit of not having all devices go through both namespace pipelines. If we are only going to pick one ...

1) **Ignore happi information**
What is the point of the metadata fields in happi then? What if there are other applications that rely on a stand name being correct without guessing it from the name? 

2) **Ignore name information**
Devices from experiment files, beamline.py won't be included in namespaces. This is a bummer.

I'd like to propose a slightly different solution. The fact that these valves didn't show up in the correct place is a bug in the device configuration, one that we shouldn't try and resolve in this library. What if we had two separate pipelines for creating namespaces; devices from happi get added based on their metadata, devices from other sources get added by their name. This means we have two small manageable functions but don't loose out on features.

ZLLentz:
Your slightly different solution is how it already works, but with a refactor.

If we do it that way, I'd suggest adding a new field in addition to beamline so we can group on "room" or "hutch" instead of path segment. A lot of the time this will be triple redundant (hutch=mfx, beamline=mfx, name=mfx_dg3_valve), but sometimes this will fix the sorting (hutch=xrt, beamline=hxd, name=xrt_dg3_valve).

I'd counter though and say that these fields are fundamentally redundant because we need to have a consistent naming scheme for devices, which should always match this metadata.

ZLLentz:
Also, I think it's confusing for the user to ever have an inconsistency between the `underscore_name` and the `underscore.name`. Doing it all through the name means there is never a namespace inconsistency, there is never an `mfx.dg3.specialname` that corresponds to a `specialname` object instead of a `mfx_dg3_specialname` object.

This is really just a browser to help the user find the object they want as soon as possible, not a sorting to help construct screens, etc.

teddyrendahl:
Yeah. I just worry that we should be utilizing our database more efficiently so the string parsing isn't requisite. Maybe we need a `happi` utility that syncs the database with the parsed string or something and at least shows a warning if you didn't enter information like `stand` that appears to be in the name.

I see your `specialname` argument and it makes sense for this application. I just don't want the sorting in the hutch python to not be consistent with our eventual screens, but I think in this case having a few extra devices in the namespaces aren't the end of the world.

ZLLentz:
Perhaps what we should move towards is a convention for the naming, where the "name" is generated automatically from the metadata fields, e.g. "{hutch}_{stand}_{class}{num}" and bake this into the database structure. Objects without all the fields would have no name and be forced inactive.

PR #120, opened 2018-04-26
FIX: disable jedi in ipython shell

ZLLentz:
## Motivation and Context

Weird bug where the jedi completion gives an extra `mro` function for `SimpleNamespace` instances... so everytime you do `namespace.<tab>` you get a bogus `mro` that isn't actually an attribute of the class. I expect someone is being too zealous in including class-only methods.

This quick shell mod restores normal completion behavior.

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -133,6 +133,8 @@ def hutch_ipython_embed(stack_offset=0):
     # Add our Bug Reporting Magic
     logger.debug('Registering bug_report magics')
     shell.register_magics(BugMagics)
+    # Disable jedi completion, it's buggy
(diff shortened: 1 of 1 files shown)


ZLLentz:
`jedi` is an autocompletion library. It's not clear to me what the extra features are over the built-in completion, I didn't notice any ill effects when I was testing this.

ZLLentz:
It looks like IPython half-uses jedi, and is waiting on jedi bugfixes to 100% it.

ZLLentz:
I'll keep on the lookout if people have issues, but I'm annoyed enough by the extra mro attribute that I'm eager to merge this without looking too deeply

PR #123, opened 2018-05-04
ENH: Camviewer Cams

### LINKED ISSUE #117: Camera Objects from Camviewer Config
## Expected Behavior

Check the camviewer config for the hutch and load the relevant area detector objects

## Context

Requested by Silke and Jason

ZLLentz:
## Description

Load camviewer.cfg file and create detector objects for the hutch

## Motivation and Context

closes #117 

## How Has This Been Tested?

Added tests

Small note in docs and docstrings

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -17,7 +17,8 @@ from pcdsdevices.mv_interface import setup_preset_paths
 
 from . import plan_defaults
 from .cache import LoadCache
-from .constants import VALID_KEYS
+from .cam_load import read_camviewer_cfg
+from .constants import VALID_KEYS, CAMVIEWER_CFG
 from .exp_load import get_exp_objs
 from .happi import get_happi_objs, get_lightpath
 from .namespace import class_namespace, tree_namespace
@@ -76,8 +77,12 @@ def load_conf(conf, hutch_dir=None):
     - Create a ``daq`` object with ``RE`` registered, using ``daq_platform``
       to define the ``platform`` argument if provided. The default value if
       ``daq_platform`` was not defined is 0.
+    - Use ``hutch`` and ``daq_platform`` keys to create the ``elog`` object
+      and configure it to match the correct experiment.
     - Use ``db`` key to load devices from the ``happi`` beamline database
       and create a ``hutch_beampath`` object from ``lightpath``
+    - Use ``hutch`` key to load detector objects from the ``camviewer``
+      configuration file.
     - Use ``load`` key to bring up the user's ``beamline`` modules
     - Use ``experiment`` key to select the current experiment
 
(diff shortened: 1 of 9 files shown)


ZLLentz:
This is stalled pending the issues raised in https://github.com/pcdshub/pcdsdevices/pull/225 getting resolved.

ZLLentz:
This is done now @teddyrendahl! Reads the camviewer config file and makes all the cameras. It makes them in parallel using a threadpool because cameras that fail from timeouts will fail in series otherwise. You may consider a future upgrade to happi that does the same thing, but with our current no-get-on-init rules this only is really needed with the areadetector stuff. Is camviewer config reading a thing that should make it into happi? Or is this fine as a semi-permanent home for the code?

teddyrendahl:
Hmm. A very good question. Honestly, in my head, this flow of information should be reversed. The camviewer configuration should automatically be generated by `happi`. This isn't the world we live in today so I think adding this PR to `hutch-python` is fine. For L2S-I, when we have a controls system that intimately needs `happi` to function we can re-route the flow of information. By that time however, the camviewer will look very different so I don't think it is worth splitting hairs about now.

ZLLentz:
Yes. This has no way of knowing what cameras are already in the environment, since there is no signal registry.

ZLLentz:
Note to self: revisit and make sure we at least don't create two of the same device ourself! This happens in the XPP file because there is a slow-rate and fast-rate entry for a camera in the cfg.

ZLLentz:
@teddyrendahl this is ready for final review

hutch_python/load_conf.py
```diff
@@ -222,6 +228,18 @@ def load_conf(conf, hutch_dir=None):
...
+        with safe_load('elog'):
+            # Use the fact if we we used the default_platform or not to decide
+            # whether we are in a specialty station or not
+            if default_platform:
+                logger.debug("Using primary experiment ELog")
+                kwargs = dict()
+            else:
+                logger.info("Configuring ELog to post to secondary experiment")
+                kwargs = {'station': '1'}
+            cache(elog=HutchELog.from_conf(hutch.upper(), **kwargs))
```
teddyrendahl: What is this?

ZLLentz: - moved elog earlier in the creation process for organization sake
- skip elog if no hutch e.g. a raw `hutch-python` call

teddyrendahl: Cool.

teddyrendahl:
@ZLLentz I just realized that if you used the `happi.loader` we would catch that you are loading the same object twice. You however have to then go `cfg` file -> `happi.Container` -> `from_container(...)`.

Worth the complication?

ZLLentz:
It's worth the complication so long as we can multithread the happi loads. This cam load process is painful single-threaded because the area detector cameras need to connect to pvs one at a time, complete with timeouts.

PR #126, opened 2018-05-23
FIX: Use the normal erroring mechanism for ImportError inside the file

### LINKED ISSUE #125: Ambiguous Error Message in Experiment File Loading
Ran into this issue that was slightly confusing. I figured it out quickly but I think a novice might be overwhelmed. If your experiment file raises an `ImportError` when it is loaded the error message generated claims that the experiment file is missing. It would be nice to disambiguate between the two cases; there is no experiment file, and you gave me an experiment file that raises an `ImportError`. 

https://github.com/pcdshub/hutch-python/blob/695c6b75e1b8ecda4a32bccb66d3311363fa6532/hutch_python/exp_load.py#L34-L38

ZLLentz:
## Description

Parse the `ImportError` for `module_name`

## Motivation and Context

closes #125 

Reviewing this code snippet, I wonder if the experiment file should use a different logging mechanism for errors. Perhaps this is the one place where the user wants to see the traceback? Perhaps also in beamline.py?

### DIFF
--- a/hutch_python/exp_load.py
+++ b/hutch_python/exp_load.py
@@ -32,11 +32,13 @@ def get_exp_objs(proposal, run):
     logger.debug('get_exp_objs(%s, %s)', proposal, run)
     expname = proposal.lower() + str(run)
     module_name = 'experiments.' + expname
-    try:
-        module = import_module(module_name)
-    except ImportError:
-        logger.info('Skip missing experiment file %s.py', expname)
-        return SimpleNamespace()
     with safe_load(expname):
-        return module.User()
+        try:
+            module = import_module(module_name)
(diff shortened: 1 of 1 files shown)


teddyrendahl:
This doesn't look like it is has a `safe_load` around it in `load_conf` are we sure it should raise?

ZLLentz:
It now has a `safe_load` in itself. Should I create an issue for refactoring so that all the `safe_load` blocks are in `load_conf`? I know we have them spread out right now.

teddyrendahl:
Oh I missed that. Yeah might be nice to make some concrete rules. `Plugins` raise errors `load_conf` catches them.

PR #128, opened 2018-05-23
FIX: Duplicate names in tree_namespace

ZLLentz:
## Description

If we're constructing the tree namespace and we encounter a leaf that already exists, we need to handle it properly. Currently it can fail and break the tree. This makes it show a warning message about two devices having a naming conflict.

## Motivation and Context

xcspython load has a conflict between `xcs_lodcm` and `xcs_lodcm_yag`

## How Has This Been Tested?

I have added a test.

### DIFF
--- a/hutch_python/namespace.py
+++ b/hutch_python/namespace.py
@@ -148,6 +148,10 @@ def tree_namespace(scope=None):
                 if not hasattr(upper_space, key):
                     setattr(upper_space, key, IterableNamespace())
                 upper_space = getattr(upper_space, key)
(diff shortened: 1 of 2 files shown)


ZLLentz:
@teddyrendahl thoughts on if this should fail silently or noisily?

ZLLentz:
I have opted to make this a WARNING message about two devices having a naming conflict. You can proceed into a session as normal, but I don't make any guarantees about the tree namespace being well-formed if there is a naming conflict.

PR #132, opened 2018-05-31
MAINT: Do not install matplotlib backend without DISPLAY

teddyrendahl:
## Description

Check that we have a `DISPLAY` environment variable before configuring the `matplotlib` backend for `IPython`. This destroys the entire session which is strange. We should allow users without displays to access non-graphics functions. A warning is printed to the terminal that alerts the user they don't have a configured `DISPLAY` so they will be aware something is strange.

Just for fun, I tried seeing what would happen if I ran a `bluesky` scan with `BestEffortCallback` to see if it exploded without `matplotlib`. This seems to error nicely and print an error that is consistent with our warning.

```python
RuntimeError: Invalid DISPLAY variable
```

We could be extra nice and not configure the `BEC` for non `display` users but this seems extreme to me.

## Motivation and Context

As a fix for https://github.com/pcdshub/Bug-Reports/issues/6

## How Has This Been Tested?

Tested locally without X-Forwarding. Still confused how this function works in Travis

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -11,6 +11,9 @@ Bugfixes
   a missing experiment file. (#126)
 - Prevent duplicate names in `tree_namespace` from breaking the tree.
   Show a relevant warning message. (#128)
+- Do not configure the ``matplotlib`` backend for IPython if a user does not
+  have a valid ``$DISPLAY`` environment variable. The most common case of this
+  is if X-Forwarding is disabled. (#132)
 
 v0.5.0 (2018-05-08)
 ===================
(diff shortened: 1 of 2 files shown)


teddyrendahl:
Looks like test just calls this function and checks for `OSError`. Where does that get raised?

ZLLentz:
I would have expected travis to fail on https://github.com/pcdshub/hutch-python/blob/0c0c897b313f9dbad46cc091c449ab86d37cfc9f/hutch_python/tests/test_tstpython.py#L28-L29 with a `CalledProcessError`

ZLLentz:
There are two bizarre things with the test that didn't fail:
- `subprocess.Popen` and affiliates behave subtly different than running scripts in a shell or with `os.system`
- `ipython` shells that get stdin passed into them before entering the shell never reach the xqcb error and exit early

I looked into adding a test that fails before this PR and succeeds after, but the test is awkward and seems to subtly ruin the session after running. Hopefully we won't hit similar issues...

PR #137, opened 2018-07-05
ENH: Scan pvs

### LINKED ISSUE #90: Scan PVs
Port the scan PVs feature from the old hutch python in a sane way. This was a feature where at each step of the scan we would put the values and the step number to dummy EPICS pvs so that it is clear in the DAQ what the scan state is. This has been requested several times as a crucial feature.

This needs to be optional and configurable on the `RunEngine` independent of any individual plan, so that means it needs to either be a preprocessor or a callback. It's unclear to me which module this should live in.

### LINKED ISSUE #131: Inaccurate Log Message
"Missing hutch from conf. Will skip DAQ." is shown when `hutch-python` is called with no config file, but a daq is loaded anyway.

ZLLentz:
## Description

Import and include. Leave disabled to avoid two sessions fighting over the pvs.

## Motivation and Context

closes #90 
closes #131 

## How Has This Been Tested?

Added include test, made sure it connects for XCS

Would you believe I forgot? I guess I'll go back and add it.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - pyfiglet
     - happi >=1.1.1
     - pcdsdevices >=0.6.0
-    - pcdsdaq >1.2.0
+    - pcdsdaq >=2.0.0
     - psdm_qs_cli >=0.2.0
     - lightpath >=0.3.0
     - elog
--- a/docs/source/daq.rst
+++ b/docs/source/daq.rst
@@ -1,0 +1,90 @@
+DAQ
+===
+
+The following objects are provided for interacting with the data acquisition
+system:
+
+    - ``daq``: controls execution of the run
+    - ``scan_pvs``: controls the auxilliary scan pvs, which are used to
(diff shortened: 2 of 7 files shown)


teddyrendahl:
Any reason we can't just put this as a device in `happi`?

ZLLentz:
I opted to add an entire daq section to the hutch-python docs, with a brief summary and a link to the full pcdsdaq docs

teddyrendahl:
Can I just pose a different course? Load the object automatically and then just have scans that subscribe it automatically. Might be nicer than adding this other `enable` /`disable` route.

ZLLentz:
Is this different than what is implemented?

teddyrendahl:
Is that what "disabled" means?

ZLLentz:
Disable and enable are shortcuts that call subscribe using the run engine

ZLLentz:
We should still clear this up here, as I wrote it into the `hutch-python` docs in this PR. I was comfortable doing it this way because 99% of the time this will be configured on a per-session basis.

PR #141, opened 2018-07-20
ENH: exp cli arg

### LINKED ISSUE #140: Command-line arg for forcing experiment
## Expected Behavior

hutch-python --exp cxics0116

This is marked as a bug because there's no way to force experiment in a 2-daq situation.

ZLLentz:
## Description

`hutch-python --exp cxilp4517` or `hutch-python --exp lp4517` can be used to override the experiment for a single hutch python session.

## Motivation and Context

closes #140 

## How Has This Been Tested?

Added tests and tried it for xpp

Added to the docs appropriately. It will also get into the `--help` string and the `--help` string in the docs automatically.

### DIFF
--- a/hutch_python/tests/test_questionnaire.py
+++ b/hutch_python/tests/test_questionnaire.py
@@ -17,6 +17,7 @@ def clear_happi_cache():
 def test_qs_load():
     logger.debug('test_qs_load')
     hutch_python.qs_load.QSBackend = QSBackend
+    clear_happi_cache()
     objs = get_qs_objs('LR12', '15')
     assert objs['inj_x'].run == '15'
     assert objs['inj_x'].proposal == 'LR12'
--- a/docs/source/experiment.rst
+++ b/docs/source/experiment.rst
@@ -35,6 +35,12 @@ experiment other than the current experiment, you can override this as:
 
 See `yaml_files` for more information.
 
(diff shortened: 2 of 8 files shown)


hutch_python/tests/test_questionnaire.py
```diff
@@ -17,6 +17,7 @@ def clear_happi_cache():
 def test_qs_load():
     logger.debug('test_qs_load')
     hutch_python.qs_load.QSBackend = QSBackend
+    clear_happi_cache()
```
ZLLentz: pre-empting the question: clearing the cache here removed a red herring of a bug, helping my debugging process

teddyrendahl: Interesting. A bug in the `happi_cache`?

ZLLentz: the happi_cache behaves as intended, the problem is that we're doing happi loads multiple times in one session, so if something is loaded incorrectly in one test but not in another, it may show up as an error in both of them- or the reverse, a bug might get masked!

teddyrendahl: Make sense. There is a way to force the method to create a new device and overwrite the cache, but there is not a global option to do so. Wonder if there should be 🤔 ?

PR #144, opened 2018-08-11
ENH: Update for new QS

ZLLentz:
See https://github.com/pcdshub/happi/pull/68

I still need to edit the docs about changes to the conf.yaml experiment override.

This will of course not work until the above PR is finished.


teddyrendahl:
Can we kick this along? Should certainly be included in latest release

PR #159, opened 2018-11-01
FIX: Fix tree_namespace

### LINKED ISSUE #149: Namespace Creation Creates Invalid Names
Related to #147 but slightly different. If a device is named `mfx_gige_01`, we place this in a namespace accessed by `mfx.gige.01` the only issue is this is an invalid Python attribute name. 

I think this requires a different fix than #147. That issue has to with an invalid input into the system, this takes a valid Python identifier and deconstructs it until it is no longer usable.

### LINKED ISSUE #150: Namespace Creation overwrites objects
There is an issue with namespace creation where we have the potential to overwrite important objects. For instance `daq_scan` creates a namespace `daq` which overwrites the `pcdsdaq.Daq` object. We sure we either do not create the namespace or we tack on the objects as to not lose any devices.

ZLLentz:
## Description

Fix a bug in this very dangerous feature... I rewrote `tree_namespace` to prevent objects from disappearing and being overwritten.

## Motivation and Context

Previously, you could define something like `daq_repeat` in `beamline.py` and it would end up creating a new tree namespace named `daq` in the main environment, overwriting the real `daq` object. Now, we will graft `daq_repeat` onto the real `daq` object instead of overwriting it.

closes #149 
closes #150 

## How Has This Been Tested?

Added tests that failed before the change


klauer:
Per our discussion: should add checks for valid attribute names (i.e., in `test_123abc` `123abc` is invalid), and clashes with builtins (specifically `in` and `for` seem likely)

hutch_python/namespace.py
```diff
@@ -131,28 +132,58 @@ def tree_namespace(scope=None):
...
+        for key, value in node.items():
+            if hasattr(obj, key):
+                # This means we would be overwriting something important
+                logger.warning('Issue in tree_namespace, tried to overrwrite'
+                               '%s.%s=%s', obj, key, getattr(obj, key))
             else:
-                setattr(upper_space, name, obj)
+                sub_node = unpack_node(value)
+                try:
+                    setattr(obj, key, sub_node)
```
teddyrendahl: I'm kind of against tacking things on here. What if it overwrites an existing attribute? 

```python
motor = EpicsMotor(...)

def motor_timeout(...):
    pass
```

teddyrendahl: I would only do this if it is an `IterableNamespace`

hutch_python/tests/test_namespace.py
```diff
@@ -66,9 +66,31 @@ def test_tree_namespace():
...
+    ns = tree_namespace(scope=scope)
+    assert ns.daq.name == 'daq'
+    assert ns.daq.repeat.echo == 'daq_repeat_echo'
+    assert ns.daq.repeat.name == 'daq_repeat'
+
+
+def test_invalid_names():
+    logger.debug('test_invalid_names')
+    # Do not crash, but cover warnings and verify not in tree
+    scope = SimpleNamespace(cheese_in_wheel=4, why_1=1)
```
teddyrendahl: Wait is the behavior we want? In my mind both these objects should be in the tree as:

```python
cheese.in_wheel
why_1
```

ZLLentz: Fair point, I'll look at it.

hutch_python/namespace.py
ZLLentz: I check if it is an existing attribute a few lines up.

hutch_python/tests/test_namespace.py
ZLLentz: After some thought, we'd have to make some rules about when to merge the invalid sections with the prior section vs the following section. The example from the test is complicated when you add additional objects e.g.:
`cheese_in`, `cheese`, `cheese_in_wheel_1`
So how do you sort these into the tree?
`cheese.in_wheel`
`cheese_in`
`cheese`
`cheese.in_wheel_1`
It falls apart, even if you have strict rules e.g. "always merge with the previous key"... you have to ask the user to change their names or accept a malformed tree.

If we're ok with a malformed tree, I will enact the "always merge with previous key" rule, so these become:
`cheese_in.wheel`
`cheese_in`
`cheese`
`cheese_in.wheel_1`
The "malformed" here is that there are a lot of `cheese` names that don't sort under `cheese`

teddyrendahl:
I'm all for the limitation. 

1. Only enforce names where we expect there to be patterns. I think this leads to less confusion in the end.
2. Right now if I'm not mistaken the grouping is the longest part of the loading process. Limiting the "seeds" of our tree should drastically lower this.

hutch_python/tests/test_namespace.py
teddyrendahl: Agreed. I think not having them might still be better then malforming them as long as they are accessible. That being said, less concerned about this if we just limit to specific groups. 

What's wrong with merging to next key then back if not available? I think that gives you:

`cheese.in_wheel`
`cheese`
`cheese_in`
`cheese.in_wheel_1`

which at least gives you three of the four and `cheese_in` will never be "splittable" because "in" will never be valid.

hutch_python/namespace.py
```diff
@@ -131,28 +132,58 @@ def tree_namespace(scope=None):
...
+                curr_dict = curr_dict[key]
+            curr_dict['_obj'] = obj
+
+    # Unpack tree_dict to make tree_space
+    def unpack_node(node):
+        obj = node.pop('_obj', None)
+        if obj is None:
+            obj = IterableNamespace()
+        for key, value in node.items():
+            if hasattr(obj, key):
```
teddyrendahl: 😄

hutch_python/tests/test_namespace.py
ZLLentz: The algorithm in my head for "next, else back" is much uglier than the one for "always previous", which is much uglier than the code as implemented. I'm a sucker for keeping code readable.

ZLLentz:
1. Agreed on less confusion
2. I did some profiling, this PR accidentally speeds up the loading considerably! (factor of 2). But it's still pretty slow (about 3s for XPP). It seems like a lot of time is being wasted by running it on "everything". If we can target the "eligible" objects in an efficient way it should drop to negligible time spent making the group.

I'm proposing the following groups, based on existing PVs:
fee, amo, sxr, hx2, xpp, hxx, pbt, xrt, xcs, cxi, mfx, mec, las

Anything obvious I missed?

teddyrendahl:
Want to group `daq` stuff together?

Also, you could only run it on `happi` objects and stuff imported from the local hutch module 🤷‍♂️ . Not sure how much this would leave on the floor.

ZLLentz:
Actually the slowdown is due to some mistake of mine in the implementation, I will fix and show you. Happy to include `daq` stuff, so maybe also `det`?

I was going to accumulate objects that start with `<groupname>_` into a separate dict during the normal loading, I think that's the fastest way to go about this. Otherwise I have to iterate through the whole list of loaded objects (or some subset) again near the end.

ZLLentz:
Ok. I convinced myself that I was doing something wrong here that made the namespace assemble super slowly. After banging my head against it for a while I finally realized something really basic and dumb:

These `debug` statements totally destroy the load speed if they are really long, but only on our pcds machines. It's just NFS punching us in the face.

Runs great now.

hutch_python/namespace.py
```diff
@@ -121,38 +122,66 @@ def tree_namespace(scope=None):
...
+        keys = name.split('_')
+        valid_keys = True
+        for key in keys:
+            if iskeyword(key):
+                logger.warning('Issue in tree_namespace, %s in %s is a '
+                               'reserved python keyword, omitting from tree',
+                               key, name)
+                valid_keys = False
+                break
+            if not key.isidentifier():
```
klauer:
```suggestion
            elif not key.isidentifier():
```

hutch_python/namespace.py
```diff
@@ -121,38 +122,66 @@ def tree_namespace(scope=None):
...
+
+    # Unpack tree_dict to make tree_space
+    def unpack_node(node):
+        obj = node.pop('_obj', None)
+        if obj is None:
+            obj = IterableNamespace()
+        for key, value in node.items():
+            if hasattr(obj, key):
+                # This means we would be overwriting something important
+                logger.warning('Issue in tree_namespace, tried to overrwrite'
```
klauer:
```suggestion
                logger.warning('Issue in tree_namespace, tried to overwrite'
```

hutch_python/tests/test_namespace.py
```diff
@@ -66,9 +66,31 @@ def test_tree_namespace():
...
+    # There was a bug where names could conflict with eachother
+    objs = dict(daq=Signal(name='daq'),
+                daq_repeat_echo='daq_repeat_echo',
+                daq_repeat=Signal(name='daq_repeat'),
+                daq_deep='daq_deep',
+                daq_deep_deep=Signal(name='daq_deep_deep'))
+    ns = tree_namespace(objs=objs)
+    assert ns.daq.name == 'daq'
+    assert ns.daq.repeat.echo == 'daq_repeat_echo'
+    assert ns.daq.repeat.name == 'daq_repeat'
```
klauer: I rather like the tree concept in general, but this portion scares me a bit. Effectively, since `daq` exists in the top-level namespace, we are monkey-patching on additional attributes (`repeat` and `deep` here).

If I call `tree_namespace`, I wouldn't expect it to be modifying _any_ of the objects I passed, merely put them in a logical grouping. At the very least, that should be well-documented.

How common is this scenario where you have `obj` and many similarly named `obj_{...}` in a single environment?

ZLLentz: I hope this is rare. I also do not like the patching, but I don't *really* know what else to do about it. Ideal scenarios would be one of the following:
1. Names are very consistent (e.g. all_with_four_sections) and we error/warn instead of monkey-patching
2. There is some way to make a proxy object tree without doing any setattrs on the real objects

I'm at a bit of a loss of how to do this correctly

ZLLentz:
I put some more thought into this, and now I think the following is the desired behavior:
- Make the tree, but only use namespaces for the nodes as Ken was saying. Objects can only be external leafs of the tree.
- If an object would have become a node, instead omit it and log in the debug. This indicates that the tree is not desirable for this object.
- If a tree group would have overwritten an object in the final namespace, drop it in favor of the original object. This prevents import things from being overwritten.
- If a tree group has only one object, drop it.
- Remove the concept of allowed groups. Just include anything we can. This will be fast now that I found the bottleneck.
- Implement the more complicated handling of fixing invalid names that Teddy was talking about.

teddyrendahl:
> If an object would have become a node, instead omit it and log in the debug. This indicates that the tree is not desirable for this object.

What is the current behavior? It sounds confusing that `mfx_dg1_pim` would be a different than `mfx.dg1.pim` if `mfx_dg1_pim_x` exists...

ZLLentz:
Current behavior (pre this PR) depends on the order the old algorithm reaches the tree. One or the other will be omitted. I'm suggesting that we should curate names so instead of `mfx_dg1_pim` and `mfx_dg1_pim_x` we have `mfx_dg1_pim_states` and `mfx_dg1_pim_x`.

klauer:
Looks like there's still work to be done here, so we should hold off on this one.

ZLLentz:
Yes, do not merge this. This "feature" is a candidate for refactor/simplification (and isn't really a rush)

ZLLentz:
closing stale

Issue #86, opened 2018-03-15
Epics Archiver Inclusion


ZLLentz:
Perhaps include an https://github.com/slaclab/archapp object

ZLLentz:
- [ ]  move archapp to pcdshub
- [ ]  set up conda recipe, travis, repo structure
- [ ]  build archapp for our conda channel
- [ ]  import the interactive client into `load_conf.py`

ZLLentz:
closed by #176

PR #176, opened 2020-01-11
Stuff archapp into load_conf.py

ZryletTC:
## Description
Loaded and cached archapp in load_conf.py to run in hutch python.

## Motivation and Context
Add the EpicsArchiver interface to new beamline python.
Addresses #86 

## How Has This Been Tested?
@ZLLentz  is testing it now.

Labeled it in load_conf.py


ZLLentz:
This worked in a quick test on xcs-control

PR #178, opened 2020-01-25
ENH: update cookiecutter

ZLLentz:
## Description

- update version on default env
- switch set-env to ssh


ZLLentz:
This needs #179 to merge before it starts working

PR #183, opened 2020-03-19
MNT: happi API changes

ZLLentz:
## Description

We're generalizing and cleaning up `happi`, so some API is changing. This adjusts `hutch-python` for it.

## How Has This Been Tested?

Tested locally with dev builds until a release of happi + lightpath

attn @klauer @hhslepicka 

There might be some need for a `DummyBackend` for stuff like the conftest hack that failed here

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,7 +22,7 @@ requirements:
     - pyyaml
     - coloredlogs
     - pyfiglet
-    - happi >1.1.1
+    - happi >=1.5.0
     - pcdsdevices >=2.3.0
     - pcdsdaq >=2.0.0
     - psdm_qs_cli >=0.2.2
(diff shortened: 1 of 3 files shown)


klauer:
What about instead of a dummy backend, we have a temporary configuration fixture that uses the json backend?

ZLLentz:
That seems good to me

ZLLentz:
I will let this sit here until we finish the happi refactors

klauer:
Looks like pcdsdevices now, maybe?

ZLLentz:
Yep, it's definitely (at least partly) in need of a `pcdsdevices` tag

ZLLentz:
I'll dig into the rest. I'd like all the packages to work together by the end of the week...

ZLLentz:
Ah. This depends on lightpath. I'll make a tag there before assuming there's a rabbit hole to enter.

klauer:
Makes sense!

PR #185, opened 2020-04-15
ENH: use pcdsutils to configure logstash logger

ZLLentz:
## Description

Set up this logger at the application level for any user code that sends messages to logstash

## Motivation and Context

Logging configuration is the responsibility of the application to maximize library portability

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - happi >1.1.1
     - pcdsdevices >=2.3.0
     - pcdsdaq >=2.0.0
+    - pcdsutils >=0.1.1
     - psdm_qs_cli >=0.2.2
     - lightpath >=0.3.0
     - elog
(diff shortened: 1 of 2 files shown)


klauer:
Looks like a `configure_pcds_loging` typo failing the build

ZLLentz:
I'm the king of typos today! The best at loging and the manifastest!

ZLLentz:
@klauer did I understand correctly that it is best to just call with default args and let environment variable overrides happen if needed?

klauer:
Yeah, I think that's the best way to do it as-is - if you have a better idea, though, I'm happy to push those changes into pcdsutils sooner rather than later.

ZLLentz:
Thinking about it a bit more, I think that environment variables are unambiguously the correct way to configure things like servers, ports, etc., because we do not want to go around adding lots of the same arguments to every application.

typhos --logstash-server --logstash-port --logstash-stash-stash
hutch-python --logstash-stash-stash
etc.

PR #195, opened 2020-05-21
MNT: Fix the docs build to fix the tag

ZLLentz:
## Description

- Build docs even if we don't upload
- Put `conda-forge` first for better results
- Pin latest of in-house dependencies

## Motivation and Context

- The tag failed because the docs don't build.
- The docs failure was missed because they weren't being built except on tag.
- The docs failed because they were running with old versions of libraries.

Truthfully there is more to be done here but this was good enough to fix the problem for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
   - conda update -q conda conda-build
   - conda config --add channels pcds-tag
   - conda config --add channels $PCDS_CHANNEL
-  - conda config --append channels conda-forge
+  - conda config --add channels conda-forge
   # Useful for debugging any issues with conda
   - conda info -a
   # Test conda build
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,10 +70,13 @@ script:
   # Build docs
   - set -e
   - |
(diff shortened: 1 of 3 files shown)


klauer:
We'll also blanket apply new CI configs on our repositories soon, which will let us fix at least some of the pain points here in one spot instead of `<total repo count here>` places... 🤞

.travis.yml
```diff
@@ -45,7 +46,7 @@ install:
   - conda update -q conda conda-build
   - conda config --add channels pcds-tag
   - conda config --add channels $PCDS_CHANNEL
-  - conda config --append channels conda-forge
+  - conda config --add channels conda-forge
```
ZryletTC, line 49: Why is this necessary? I'm kind of worried about reprecussions

ZLLentz: technically not necessary in this PR, but it was the only difference between my local config that created the correct environment and the travis config here that created the wrong environment

ZryletTC: Huh ok, conda is weird :(

ZLLentz: The solver needs to make some sort of default judgement if it has multiple mutually exclusive solutions for the environment. Since the default judgement can depend on channel order and priority, it can get messy. We should avoid less than and equal to pins at all costs in our dependencies to minimize these cases of multiple incompatible solutions.

PR #198, opened 2020-05-29
TST: Cleanup Freeze Workaround

ZLLentz:
## Description

Pyepics/ophyd combination currently freezes at close due to a pyepics bug. This makes tests hang at the end if they don't do dumb monkeypatches.

## Motivation and Context

The test hanging has a knock-on effect where it causes the test to fail the "big" pcds-envs test. This is because pcds-envs assumes it will take less than 5 minutes to run each test suite. This is reasonable because there are many packages to test, and if they take forever to run it will overrun the max duration of a travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,12 +62,6 @@ script:
   - coverage report -m
   - flake8 hutch_python
   - flake8 bin
-  # Test again, with the installed version
-  - mkdir notest
-  - mv hutch_python/*.* notest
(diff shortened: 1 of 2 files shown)


ZLLentz:
This did lead to an improvement on the pcds-envs test script though. Eventually a test was going to time out and leave me wondering wtf happened in the big build output.

klauer:
Silver linings!

PR #199, opened 2020-07-01
MNT: pack camviewer into namespace to avoid name conflicts

ZLLentz:
## Description

Rather than pass every camviewer area detector into the user's ipython shell, collect them into a `SimpleNamespace`

## Motivation and Context

This avoids name conflicts between devices defined in camviewer and devices defined in happi

## How Has This Been Tested?

Worked at XPP, almost certainly busts this repo's test suite.

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -7,6 +7,7 @@ import yaml
 from copy import copy
 from pathlib import Path
 from socket import gethostname
+from types import SimpleNamespace
 
 from bluesky import RunEngine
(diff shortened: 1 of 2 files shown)


ZLLentz:
The resulting name space also does not work with tab-completion... it seems that when half the contents of the namespace raise exceptions on getattr this totally fails.

ZLLentz:
It actually might be the reprs causing the tabs to blow up... I kind of want to investigate but kind of feel like this will be a rabbit hole.

klauer:
Hmm... Seems to be a nicety, if you don't have time for the likely rabbit hole.

PR #203, opened 2020-07-24
TST: update tests to match lightpath update

ZLLentz:


ZLLentz:
Wow my tabs are showing, that's embarassing

PR #214, opened 2020-08-28
FIX: initialize teleporter bug and segfault once bluesky PR 1395 is m…

aegger13:
Use teleporter, set use_teleporter argument for bec

pretty much as above
## Motivation and Context

segfault trying to use matplotlib in bec on NFS machine

## How Has This Been Tested?

I ran multiple plots using bec without any errors or segfaults.  This also relies on my changes to blue sky from PR 1395 being in a released version https://github.com/bluesky/bluesky/pull/1395

Using the teleporter should be an arg/env variable or something.  This will default to always using BEC with teleporter

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -10,6 +10,7 @@ from socket import gethostname
 from types import SimpleNamespace
 
 from bluesky import RunEngine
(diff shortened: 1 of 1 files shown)


aegger13:
Different fix in PR 215, closing

Issue #210, opened 2020-08-25
Liveplot explodes in hutch-python on NFS


ZLLentz:
When doing a bluesky scan in a hutch python session, after initializing the qt teleporter, something bad happens that causes a segfault. The errors suggest that the qt teleporter isn't doing the right thing.

Everything works great in the same environment if:
- you load a normal (not hutch-python) ipython session
- you do it on my desktop machine (AFS)

Can reproduce by:
- ssh psbuild-rhel7
- source pcds_conda
- hutch-python
- set up the qt teleporter (should be done inside hutch python but it is new enough that it isn't yet)
- try to do any scan that would produce a liveplot from bec

aegger13:
tentatively assigning to myself, but will likely ask for help

ZLLentz:
Minimal example script that breaks:
```
In [1]: from ophyd.sim import motor1, det1

In [2]: RE(bp.scan([det1], motor1, -10, 10, 21))
```

ZLLentz:
There is also a new `initialize_qt_teleporter` function that needs to be called before anything works:
```
from bluesky.callbacks.mpl_plotting import initialize_qt_teleporter

initialize_qt_teleporter()
```

ZLLentz:
With the above, everything works in a fresh ipython session but fails in a hutch python session.

klauer:
May be relevant:
https://github.com/pcdshub/hutch-python/blob/d7920dcb62be21b77c477d442fa79606ba155dff/hutch_python/cli.py#L117-L118

ZLLentz:
That's how you configure ipython to load with the qt backend without requiring the user to pass %matplotlib qt5. I agree that it is one possibility for the source of the issue.

Other things to check:
- Does the timing of the qt teleporter matter? E.g. does it have to be done before BEC?
- Does this happen when using a fresh RE/BEC rather than the ones hutch-python gives you?

aegger13:
Update
I spent some time digging into this.  As a first pass, it looks like there is an issue with tk internals for BestEffortCallback not being called on main thread.  Why this happens only on NFS seems very strange.  
Teleporter doesn't seem to have any impact whatsoever. 
 I wanted to try and use 'Agg' for Matplotlib backend as it seems to deal better in threaded environments, but I get complaints about it already being set to 'Qt5Agg'.  
[...]

PR #215, opened 2020-08-31
force use of Qt5Agg backend before BEC initialization

aegger13:
Use Qt5Agg backend and initialize qt teleporter

Initialize Qt Teleporter, explicitly force Qt5Agg as backend as matplotlib import defaults to TkAgg

## Motivation and Context

segfault from use BestEffortCallback

https://github.com/pcdshub/hutch-python/issues/210

## How Has This Been Tested?

Tested that seg faults go away without Bluesky, plots show up as normal

Tested on cxi-daq, psbuild-rhel7

This should not affect any other code


aegger13:
I found that when we import matplotlib, it does use TkAgg as the backend by default.  I agree that the easiest fix is to just explicitly set the backend (I think Qt5Agg is the one we want?).  This isn't super satisfying as I don't know where the offending matplotlib import with the tk backend is coming from for sure, all mpl sources seem to use the right backend.  I can close the other PR.

klauer:
It seems to me that `import matplotlib; matplotlib.use('...')` is important to have before everything else, as we're unsure where the tk backend selection is coming in. 

A few options:
1. The flake8 errors are indicating the style is wrong, but not understanding the purpose behind it.  A ` #  noqa` at the end of the `use()` line may fix it. It may also be a good idea to add an `# isort: skip` to tell the tool we use to sort our imports to ignore this line.
2. An additional module could be introduced that just does `import; matplotlib.use()`
3. I could be wrong, but I think there's a matplotlib environment variable for the backend that could be used

ZLLentz:
Ken, there definitely is an environment variable for the matplotlib backend. I think this makes life a bit confusing though, since we've decided that the backend choice matters a lot here for the bluesky compatibility and managing the backend in a different place may cause confusion.

aegger13:
Ugh, so # noqa doesn't work (I don't know if we need to install flake8-noqa specifically for this to work, but I'm following the docs and no luck).  Sorry for bogging down Jenkins, running flake8 locally now.  I think the easiest solution is to just have a module that does this explicitly for us, where do you think I should locate this in the code base?

klauer:
No need to install `flake8-noqa` - it's a built-in thing. 

Giving it a try locally, this appeases flake8 but is arguably as ugly as it could get (especially with the `;`):
`import matplotlib; matplotlib.use('Qt5Agg')  # noqa # isort: ignore`

```
$ flake8 hutch_python
$
```

Module would probably be better, but it's up to you guys 👍

ZLLentz:
Ken, you're suggesting adding something like a `hutch_python.mpl_config` module that just imports matplotlib and sets the backend? Then in the conf loader we do `import .mpl_config`?

I think something like that would be a very clean solution

klauer:
Yes, @ZLLentz

aegger13:
The module approach leads to the F401 '.mpl_config' imported but unused error.  Should I just go with # noqa: #isort: ignore approach.  Otherwise, I think I'd need to have a function that is just a pass from the module to quite this error, which is kind of silly.  lol, PEP style guides, how I love thee.

klauer:
Yeah - `from . import mpl_config   # noqa: F401 # isort: ignore`

klauer:
Qt selection breaks docs building on Travis:
```
qt.qpa.xcb: could not connect to display 

qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.

This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

Should add `pyqt` to `dev-requirements`

ZLLentz:
@aegger13 if the CI/docs build is busted here, but all the tests pass, I am ok with merging and creating a new issue to investigate the CI failure separately. I suspect that the "could not connect to display" error could be a side-effect of this travis configuration not having the same display modules activated that we use for testing `typhos`, etc., and I also think this repo needs a big CI update anyway to bring it in line with our other modules.

ZLLentz:
Specifically, I think we need the contents of this file in the travis config:
https://github.com/pcdshub/pcds-ci-helpers/blob/master/travis/shared_configs/setup-env-ui.yml

aegger13:
All right, it does seem specific to travis env setup as it works on NFS and my local machine.  I'll stop here.  That's enough make a change, do something else, see change fail, repeat...

ZLLentz:
If it works on NFS, but fails on travis, we are in a hugely better situation than before!

ZLLentz:
Ok, I will merge this and post a new issue. I might even tackle that one today before someone asks me to do something else.

aegger13:
no problem, and that was quite an annoying issue!  Alway happy to tackle an issue that you guys don't have time for.

PR #220, opened 2020-09-10
ENH: ship uncaught exceptions off to logstash

### LINKED ISSUE #219: Integrate with the centralized logging system
Using pcdsutils, log exceptions, useful performance metrics, etc.

https://github.com/pcdshub/pcdsutils/blob/master/pcdsutils/log.py

klauer:
Closes #219 

## Description
Uncaught exceptions in a hutch-python environment will be shipped off to logstash with this PR.

## Exceptional exceptions
These would just be spammy and are ignored:
* `KeyboardInterrupt`
* `SystemExit`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: no-commit-to-branch
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-xml
+    -   id: check-yaml
+        exclude: '^(conda-recipe/meta.yaml)$'
+    -   id: debug-statements
+
+-   repo: https://gitlab.com/pycqa/flake8.git
+    rev: 3.8.3
+    hooks:
+    -   id: flake8
+
+-   repo: https://github.com/timothycrosley/isort
+    rev: 5.5.2
+    hooks:
+    -   id: isort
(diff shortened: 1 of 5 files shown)


ZLLentz:
We'll need to make sure this doesn't interfere with the existing exception log handling in https://github.com/pcdshub/hutch-python/blob/master/hutch_python/ipython_log.py

ZLLentz:
Ken, did you give this a go already? e.g.
`hutch-python`
`[1]: 1/0`

ZLLentz:
CI failure unrelated

klauer:
I just circled back to this PR this morning. 

You were right that one would conflict with the other; but it's the existing implementation that conflicted with this one. Refactoring as we speak.

ZLLentz:
I find that pretty surprising!

ZLLentz (approved):
This looks clean to me, so if you tell me the message makes it into logstash I am happy to merge

klauer:
The good:
image

The bad:
image

The ugly (double exception traceback in IPython):

image

The double-exception is what will block this, and I'm not sure what to do about it.

ZLLentz:
Is this happening automatically because the level is ERROR, and we have some logging config to send all >= INFO to the stdout stream? And then we need it to be ERROR for correctness in logstash?

ZLLentz:
Yeah this repo definitely configures the root logger... maybe we fix that first

klauer:
Hmm, that'll do it...

klauer:
Shying away from fixing our logging configuration, the log config file gave a hint to check into the 'propagate' setting. It appears to have fixed the issue:

```python
In [1]: import logging

In [2]: logger = logging.getLogger('pcds-logging')

In [3]: logger.handlers
Out[3]: [<_JsonLogQueueHandler (DEBUG)>]

In [4]: logger.propagate
Out[4]: False

In [5]: logger.info('Hi, logstash!')

In [6]: 1./0
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-6-d650afc74941> in <module>
----> 1 1./0

ZeroDivisionError: float division by zero
```

Pushed a hutch-python fix along with a pcdsutils fix (https://github.com/pcdshub/pcdsutils/pull/20).

ZLLentz:
https://docs.python.org/3/library/logging.html#logging.Logger.propagate
TIL

No comment on typo in file for default argument

ZLLentz (approved):

klauer:
CI failure unrelated as Zach mentioned above; let's go ahead with this and see how it works out.

PR #221, opened 2020-09-16
Revert "MNT: remove jedi deactivation, since it doesn't seem like an …

klauer:
…issue now"

This reverts commit 293b301b74b952eccdf3314cb64138a0b1ac4575.

## Description
Anecdotally, tab completion _appears_ faster on psbuild, though I never experienced the extreme slowness reported in https://github.com/pcdshub/Bug-Reports-and-Requests/issues/39.

## Motivation and Context
Tab completion appears to be slow in certain cases and users are unhappy.

One of those users who has experienced this slowdown should give this a try prior to merging.

May close https://github.com/pcdshub/Bug-Reports-and-Requests/issues/39

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -120,6 +120,8 @@ def main():
             logger.warning('No DISPLAY environment variable detected. '
                            'Methods that create graphics will not '
                            'function properly.')
+        # Avoid bugs, probably removable at some point
+        ipy_config.InteractiveShellApp.Completer.use_jedi = False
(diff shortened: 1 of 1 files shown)


ZLLentz (approved):
To me this is a no-brainer to merge.

ZLLentz:
The CI failure is related to CI config, and not to this PR.

ZLLentz:
Apparently "may close" means "does close" to github. I re-opened the issue, since there could be more here.

PR #226, opened 2020-10-02
MNT: disable jedi three times for good luck

ZLLentz:
## Description

Referring to https://ipython.readthedocs.io/en/stable/config/options/terminal.html, update our jedi disabling config appropriately.

## Motivation and Context

- `use_jedi` makes tab completion very slow, and it is unclear that it helps in a hutch-python session, so we don't want it.
- Extra items in the IPython config are silently ignored, so it's ok to cover for both APIs
- closes https://github.com/pcdshub/Bug-Reports-and-Requests/issues/39

## How Has This Been Tested?

Only interactively- I could not see a way to open the IPython terminal in a unit test and have it not break the test suite.

N/A

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -120,8 +120,11 @@ def main():
             logger.warning('No DISPLAY environment variable detected. '
                            'Methods that create graphics will not '
                            'function properly.')
-        # Avoid bugs, probably removable at some point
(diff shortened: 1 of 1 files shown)


klauer (approved):
_Out, darn jedi! Out, I say!_

ZLLentz:
_Only a sith deals in absolutes_

Issue #224, opened 2020-09-24
Warn if namespace names clash


klauer:

ZLLentz:
https://github.com/pcdshub/hutch-python/blob/ec84c2d99e5dd66c80fd1bf9cb3d447a6f3f8387/hutch_python/namespace.py#L152

klauer:
I think this needs to be even noisier.

Or we need something in the startup to say `"HEY! There were these warnings that probably require your attention; please take a look and ensure you understand the consequences:"`

ZLLentz:
I'm considering nuking the whole automatic namespace thing entirely... it creates very confusing situations.

I agree that if it stays, it needs to fail noisily.

PR #230, opened 2020-10-02
MNT: disable tree namespace until issues are resolved

ZLLentz:
## Description

Apply hotfix from dev/deployed hutch-python, add a comment for context.
This is the (dangerous) namespace that lets you do:
`xpp.ccm`, for example, instead of `xpp_ccm`. It creates these big cascading namespaces of devices.

## Motivation and Context

Tree namespace can cause havoc and break top-level devices
See #224, #159 

## How Has This Been Tested?

No complaints from its omission (yet) in dev/deployed hutch-python

### DIFF
--- a/hutch_python/tests/test_load_conf.py
+++ b/hutch_python/tests/test_load_conf.py
@@ -23,7 +23,7 @@ def test_file_load():
     err = '{} was overriden by a namespace'
     for elem in should_have:
         assert not isinstance(objs[elem], SimpleNamespace), err.format(elem)
-    assert 'tst' not in objs # Tree namespace should be disabled
+    assert 'tst' not in objs  # Tree namespace should be disabled
     assert len(Presets._paths) == 2
 
 
--- a/hutch_python/tests/test_load_conf.py
+++ b/hutch_python/tests/test_load_conf.py
@@ -23,7 +23,7 @@ def test_file_load():
     err = '{} was overriden by a namespace'
     for elem in should_have:
         assert not isinstance(objs[elem], SimpleNamespace), err.format(elem)
-    assert 'tst' in objs
+    assert 'tst' not in objs # Tree namespace should be disabled
     assert len(Presets._paths) == 2
 
 
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - pyyaml
     - coloredlogs
     - pyfiglet
-    - bluesky >= 1.6.0
+    - bluesky >=1.6.0
     - happi >=1.6.0
     - pcdsdevices >=2.6.0
     - pcdsdaq >=2.3.0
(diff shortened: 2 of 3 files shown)


klauer (approved):
@ZryletTC and I got bit by this recently, and it sure was a confusing one to debug.

👍 for disabling it for now - let's see if there are any complaints and then go from there.

ZLLentz:
Ah I bet this fails a test that checks for the tree

hutch_python/tests/test_load_conf.py
```diff
@@ -23,7 +23,7 @@ def test_file_load():
     err = '{} was overriden by a namespace'
     for elem in should_have:
         assert not isinstance(objs[elem], SimpleNamespace), err.format(elem)
-    assert 'tst' in objs
+    assert 'tst' not in objs # Tree namespace should be disabled
```
ZryletTC, line 26:
```suggestion
    assert 'tst' not in objs  # Tree namespace should be disabled
```

ZryletTC (approved):
👍 on disabling for now

Imports need to be commented out or noqa'd to appease the flake8 god:
```
hutch_python/load_conf.py:28:1: F401 '.namespace.tree_namespace' imported but unused
hutch_python/load_conf.py:31:1: F401 '.utils.count_ns_leaves' imported but unused
```

ZLLentz:
There's something wrong with the travis configuration, we're installing hutch-python=v1.1.1 from pcds-tag instead of installing the version we built in the earlier build stage. I'm thinking there's some weird edge case where v1.1.1 has lighter dependencies and the ci helpers isn't using any sort of channel priority.

klauer:
Didn't notice that before ☹️

Maybe if we add the following here:
https://github.com/pcdshub/pcds-ci-helpers/blob/d1bb15ace06cfd8fdda3f5ccad0981fcc59dfbe0/travis/shared_configs/anaconda-build.yml#L17
```
- if [ ! -z "$CONDA_BUILD_CHANNEL" ]; then conda config --add channels $CONDA_BUILD_CHANNEL; fi
```

and per-repo:
```
- CONDA_BUILD_CHANNEL="file://$TRAVIS_BUILD_DIR/bld-dir"
```

I think the TRAVIS_BUILD_DIR will still be correct when using the ci scripts

ZLLentz:
Passes all tests locally and has a clean flake8
I think I merge?

klauer:
Yeah, no reason to hold up this PR on the CI issue. Will open an issue in ci-helpers.

ZLLentz:
The channel does still get added, you see it in the conda config output right before creating the env. But in this case, a random old build wins the install war.

ZLLentz:
https://github.com/pcdshub/pcds-ci-helpers/blob/d1bb15ace06cfd8fdda3f5ccad0981fcc59dfbe0/travis/shared_configs/python-tester-conda.yml#L21

PR #233, opened 2020-10-02
ENH: add class namespace for ami detectors

### LINKED ISSUE #212: Add detectors namespace
Much like we do with slits and motors, create a detectors namespace (d) that contains all ami detectors. See the pcdsdaq repo for the relevant classes to search for.

Original card:
(Tyler P):-> d.namespace (tbh idr what this is supposed to be)
(Silke):-> those might have been the 'detectors' which were Ami-detectors. Once ami is fixed a standard list of ami detectors should be added: typically this is the ipm sums and single channels of the user boxes. Not much else is typically used in the beamline python session at this point.

ZLLentz:
## Description

Add a class namespace that accumulates the AmiDet objects, as in old python

## Motivation and Context

Closes #212

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -310,6 +310,7 @@ def load_conf(conf, hutch_dir=None):
     with safe_load('default groups'):
         default_class_namespace('EpicsMotor', 'motors', cache)
         default_class_namespace('Slits', 'slits', cache)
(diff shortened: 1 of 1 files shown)


klauer (approved):

klauer:
@ZLLentz on 🔥

PR #243, opened 2020-10-21
ENH: LCLS, BeamStats, Load Order, Scan PVs

### LINKED ISSUE #152: BeamStats
Every hutch should load the `BeamStats` object. Unclear whether this should be a `happi` device at it has no configuration information.

### LINKED ISSUE #189: Scan PVs enable in config file
## Expected Behavior

Allow user to load scan pvs as enabled in the configuration

## Current Behavior

No such allowance, therefore there is always a load line for "scan pvs (disabled)

### LINKED ISSUE #241: Need to include BeamStats and LCLS classes
## Expected Behavior

Every session should have these key information classes from pcdsdevices

## Possible Solution

Add a stage to load_conf

## Context

https://github.com/pcdshub/Bug-Reports-and-Requests/issues/33

### LINKED ISSUE #242: Load questionnaire objects sooner
## Expected Behavior

qs objects should be loaded prior to beamline

## Current Behavior

qs objects loaded very late

ZLLentz:
## Description

A bunch of small additions/fixes all in one PR:
- LCLS object
- BeamStats object
- Fix load order so beamline and experiment files happen as late as possible
- Enable scan PVs for all consumers

## Motivation and Context

- closes #152
- closes #189 
- closes #241
- closes #242

## How Has This Been Tested?

TBH I have not tested this yet, one sec

### DIFF
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -8,6 +8,8 @@ from pathlib import Path
 from socket import gethostname
 from types import SimpleNamespace
 
+from . import mpl_config  # noqa: F401
+
 import yaml
 from archapp.interactive import EpicsArchive
 from bluesky import RunEngine
@@ -18,7 +20,6 @@ from pcdsdaq.daq import Daq
 from pcdsdaq.scan_vars import ScanVars
 from pcdsdevices.interface import setup_preset_paths
 
-from . import mpl_config  # noqa: F401 # isort: ignore
 from . import plan_defaults, sim
 from .cache import LoadCache
 from .cam_load import read_camviewer_cfg
--- a/hutch_python/load_conf.py
+++ b/hutch_python/load_conf.py
@@ -25,6 +25,7 @@ from .cam_load import read_camviewer_cfg
 from .constants import CAMVIEWER_CFG, VALID_KEYS
 from .exp_load import get_exp_objs
 from .happi import get_happi_objs, get_lightpath
+from .lcls import global_devices
 from .namespace import class_namespace
 from .qs_load import get_qs_objs
 from .user_load import get_user_objs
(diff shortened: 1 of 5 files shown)


hutch_python/load_conf.py
```diff
@@ -3,33 +3,33 @@
 The file's specification can be found on the `yaml_files` page.
 """
 import logging
-import yaml
 from copy import copy
 from pathlib import Path
 from socket import gethostname
 from types import SimpleNamespace
 
-from . import mpl_config  # noqa: F401 # isort: ignore
```
ZLLentz, line 12: isort is being mean...

klauer: Try bumping its version in pre-commit to latest (5.6.4) 🤞

ZLLentz: No permutations seem to work. There is a difference between where isort moves this line with or without an ignore and with or without a skip, but none of the permutations keep it where it needs to be. I'm going to look into other options for configuring isort.

klauer:
Not again... what's happening with CI? https://travis-ci.com/github/pcdshub/hutch-python/jobs/403506883

ZLLentz:
This is unrelated to the earlier issues I think

klauer:
I think you're right - looks like pcdsdevices v3.1.0 just hasn't reached conda-forge just yet.

ZLLentz:
I mean, this dependency conflict hieroglyph is suggesting that some dependencies don't support 3.6 any more. And this repo can only do 3.6 due to some of the deps that haven't been cleaned up... But I don't think the dependencies suddenly stop supporting 3.6 even in old versions overnight.

ZLLentz:
Maybe the non-conda-forge builds are just bad? And it messed up the dependency resolution?

klauer:
Re-ran after pcdsdevices new version got released - CI is back in business 🥳

hutch_python/load_conf.py
klauer: Seems to be an isort bug then, I suppose. I'll make an issue to refactor this down the line so we can still use it.

ZLLentz:
Oh thank god

ZLLentz:
@klauer any further thoughts on the PR? I think it's pretty boring/straightforward

klauer (approved):
Looks good to go, @ZLLentz!

PR #246, opened 2020-10-29
ENH: initial setup add epicsarch-qs cmd script

cristinasewell:
Create a second entry point `epicsarch-qs` that will handle creating the `epicsArch` file from the Questionnaire.

## Motivation and Context

Trying to implement and close https://github.com/pcdshub/Bug-Reports-and-Requests/issues/52

## How Has This Been Tested?

Locally only.

### DIFF
--- a/hutch_python/epics_arch.py
+++ b/hutch_python/epics_arch.py
@@ -6,6 +6,8 @@ import sys
 from happi.backends.qs_db import QSBackend
 import os
 from .constants import EPICS_ARCH_FILE_PATH
+import hutch_python.constants
+
 
 logger = logging.getLogger(__name__)
 
@@ -13,7 +15,7 @@ logger = logging.getLogger(__name__)
 parser = argparse.ArgumentParser(description='Create an epicsArch file from'
                                              ' the Quesionnaire')
 
-parser.add_argument('experiment', nargs="?", help='Experiment name to'
+parser.add_argument('experiment', help='Experiment name to'
                     ' create the epicsArch file from. E.g.: xpplv6818')
 
 parser.add_argument('--hutch', action="store",
@@ -28,20 +30,20 @@ parser.add_argument('--dry-run', action='store_true', default=False,
 
 
 def epics_arch_qs(args):
+    """Command Line for epicsarch-qs."""
     args = parser.parse_args(args)
 
     if args.experiment and not args.dry_run:
+        # set the path to write the epicsArch file to.
+        if args.path:
+            overwirte_path(args.path)
+        elif args.hutch:
+            overwirte_hutch(args.hutch)
+        else:
+            set_path(args.experiment)
         logger.info('Creating epicsArch file for experiment: %s',
                     args.experiment)
-        print('HELLO --- for now')
-        # create_file(exp_name)
-    elif args.hutch:
-        # Overwrite the default hutch
-        print(f'HUTCH----------------- {args.hutch}')
-        # EPICS_ARCH_FILE_PATH.format(hutch)
-    elif args.path:
-        # Overwrite the default path
-        print(f'PATH----------------- {args.path}')
+        create_file(args.experiment)
     elif args.dry_run:
         if args.experiment:
             print_dry_run(args.experiment)
@@ -51,6 +53,34 @@ def epics_arch_qs(args):
     else:
         parser.print_usage()
         return
+
+
+def overwirte_hutch(hutch_name):
+    """
+    Overwirte the default Hutch.
+    """
+    if hutch_name:
+        hutch_python.constants.EPICS_ARCH_FILE_PATH = (
+            EPICS_ARCH_FILE_PATH.format(hutch_name.lower())
+        )
+
+
+def overwirte_path(path):
+    """
+    Overwrite the default path.
+    """
+    if path:
+        hutch_python.constants.EPICS_ARCH_FILE_PATH = path
+
+
+def set_path(exp_name):
+    """
+    Figure out the huch name from the experiment and set the path with it.
+    """
+    if exp_name:
+        hutch_python.constants.EPICS_ARCH_FILE_PATH = (
+            EPICS_ARCH_FILE_PATH.format(exp_name[0:3])
+        )
 
 
 def print_dry_run(exp_name):
@@ -73,7 +103,7 @@ def print_dry_run(exp_name):
 
 def get_questionnaire_data(exp_name):
     """
-    Returns a list of items that would be written in the epicsArch file.
+    Return a list of items that would be written in the epicsArch file.
 
     Parameters
     ----------
@@ -87,8 +117,6 @@ def get_questionnaire_data(exp_name):
     """
     data_list = []
     qs_client = happi.Client(database=QSBackend(exp_name, use_kerberos=True))
-    # qs_client = happi.Client(
-    #    path='/home/cristina/workspace/device_config/db.json')
     items = qs_client.all_items
     if not items:
         logger.warning("No devices found in PCDS Questionnaire for %s",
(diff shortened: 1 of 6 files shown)


cristinasewell:
This is work in progress - but anybody is welcomed with early feedback

hutch_python/epics_arch.py
```diff
@@ -0,0 +1,246 @@
...
+from happi.backends.qs_db import QSBackend
+
+import hutch_python.constants
+
+from .constants import EPICS_ARCH_FILE_PATH
+
+logger = logging.getLogger(__name__)
+
+# Simple Argument Parser Setup
+parser = argparse.ArgumentParser(description='Create an epicsArch file from'
```
klauer, line 17: I'd recommend refactoring to use the following structure:

```python
def _create_parser():
    ...
    return argparse.ArgumentParser()

def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
    ...   # actually do the stuff here

def main():
    parser = _create_parser()
    parsed_args = parser.parse_args()
    kwargs = vars(parsed_args)
    create_arch_file(**kwargs)

if __name__ == '__main__':
   main()
```

1. It gives a clear `main()` entry point
2. The parser isn't global but created on-demand (and nicely tucked away in a function)
3. You can call `create_arch_file` from Python without making a list of string args
4. `main()` is small, only binding together argument parsing with calling the function that actually implements things
5. Overriding globals to implement the functionality as currently proposed is not a good thing (perhaps an anti-pattern even). Use default arguments for the function, and default arguments in the argument parser, and you sidestep these issues.

There are other things you could do to improve on what I'm suggesting, but I think it presents a reasonably good basis for a general CLI tool.

cristinasewell: Ok thanks for the feedback @klauer 
One question though:
- the `experiment` should be a required positional argument correct?, the user will not be able to do this:
change the path first:
`epicsarch-qs --path /change/the/path`
then create the file:
`epicsarch-qs experiment`

?

cristinasewell: You actually answered this already in `def create_arch_file(experiment, hutch=None, path=None, dry_run=False):` - but want to double check that this is what you meant

hutch_python/epics_arch.py
```diff
@@ -69,75 +91,20 @@ def epics_arch_qs(args):
...
-            set_path(args.experiment)
-        create_file(args.experiment)
-    elif args.hutch:
-        overwrite_hutch(args.hutch)
-    elif args.path:
-        overwrite_path(args.path)
-    elif args.dry_run:
-        if args.experiment:
-            print_dry_run(args.experiment)
+        if path:
```
cristinasewell, line 97: I am still not sure if I'm thinking correctly here about the path.

hutch_python/epics_arch.py
```diff
@@ -7,40 +7,62 @@
...
-                    ' create the epicsArch file from. E.g.: xpplv6818')
+def _create_parser():
+    """Argument Parser Setup. Define shell commands."""
+    parser = argparse.ArgumentParser(description='Create an epicsArch file '
+                                                 'from the Quesionnaire')
+
+    parser.add_argument('experiment', help='Experiment name to'
+                        ' create the epicsArch file from. E.g.: xpplv6818')
+
+    parser.add_argument('--hutch', action="store",
```
cristinasewell, line 23: So should I have some proper `default` values here for the `--hutch` and `--path`? I am not sure what they should be

hutch_python/epics_arch.py
cristinasewell: Thanks for explaining this!

klauer: No problem - hope you see the benefit in the above method. It may seem pretty simple, but I've only arrived at it after making many other badly-structured tools...

hutch_python/epics_arch.py
klauer: They should reflect the defaults you have in `create_arch_file` - `None` is correct here.

As far as the help string - the key to this parameter is that it overrides detection from the experiment name, isn't it?

hutch_python/epics_arch.py
klauer: I think it's looking better 👍

hutch_python/epics_arch.py
cristinasewell: Yeah it makes sense, you explained that to me in a different review a while ago, and I thought this time I actually got it, because I made the `create_file` "independent" - (or so I thought hahah) - but even the way I had it before it did not have all the parameters that the command line would - would not be the same.

hutch_python/epics_arch.py
cristinasewell: yes that's correct, need to update that. 
Here if the experiment is `xpplv6818` and `--hutch xcs` will write to `/cds/group/pcds/dist/pds/xcs/misc/` and the file will be called `ecpisArch_xpplv6818`

cristinasewell:
Regarding this: https://github.com/pcdshub/hutch-python/blob/af104ac320d9c91df1bd2fd63f5e33c0edf09f4b/hutch_python/epics_arch.py#L162-L170 

should I have had here something more similar to this guy: https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/hutch_python/qs_load.py#L53-L72 

where we use either `# Ws-auth `or `Kerberos` to authenticate?

hutch_python/epics_arch.py
cristinasewell: Updated this in af104ac

klauer:
I've always used the Kerberos tokens as it's easier/better than retyping your password or putting in plaintext in `qs.cfg`. I'm guessing that's probably a hook for hutch users without the ability to get a token?

Here's what I'm thinking:
1. This user/password/use-kerberos thing should really be contained in a configuration file, just like how `happi.cfg` links to `db.json` for our JSON backend
2. We probably don't want to bother fixing up happi just for this
3. To that end, perhaps user/password/use_kerberos settings should be additional options here, defaulting to `use_kerberos=True`

But before we act on this, I think roping in @ZLLentz to confirm/correct my assumptions above would be a good thing.

ZLLentz:
RE: the config file, this was set up for qs stuff because the hutch operator users don't have kerberos auth
I'll be back to this PR in a bit to take a look at the code

ZLLentz:
Ahh I see the context now!

Yes, I think we need to support both modes: qs file for the operators and kerberos for normal users. To that end it would probably be cleanest to refactor the `qs_load.py` to bring this part into it's own function and import in the epics arch writer (e.g. like a `get_qs_client` function or some equivalent name)

https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/hutch_python/qs_load.py#L53-L72

ZLLentz:
Rest of the code looks great to me

cristinasewell:
> Ahh I see the context now!
> 
> Yes, I think we need to support both modes: qs file for the operators and kerberos for normal users. To that end it would probably be cleanest to refactor the `qs_load.py` to bring this part into it's own function and import in the epics arch writer (e.g. like a `get_qs_client` function or some equivalent name)
> 
> https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/hutch_python/qs_load.py#L53-L72

I implemented this in 4481504 - please let me know if this is not what you meant and thank you.

ZLLentz:
Looks clean to me :+1:, I'll look through one more time to double-check the epicsArch writing

setup.py
```diff
@@ -10,5 +10,7 @@
       include_package_data=True,
       description=('Launcher and Config Reader for '
                    'LCLS Interactive IPython Sessions'),
-      scripts=['bin/hutch-python']
+      entry_points={'console_scripts': [
+            'hutch-python=hutch_python.cli:main',
```
ZLLentz, line 14: thanks for sneakily fixing this for me

ZLLentz:
The last thing I can think of is adding another line to the recipe test like this:
https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/conda-recipe/meta.yaml#L45

So that if for some reason the entry point breaks entirely (e.g. someone typos in the setup.py file) it will be caught when the tests try to conda build the package

cristinasewell:
> The last thing I can think of is adding another line to the recipe test like this:
> https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/conda-recipe/meta.yaml#L45
> 
> So that if for some reason the entry point breaks entirely (e.g. someone typos in the setup.py file) it will be caught when the tests try to conda build the package

@ZLLentz  is this what you meant? bbc494d

ZLLentz:
Yes, exactly. So now we see travis happily find the entrypoint after a conda install: https://travis-ci.com/github/pcdshub/hutch-python/jobs/425842578#L3199-L3212

ZLLentz (approved):

ZLLentz:
For whatever reason I forgot to press the merge button here

PR #255, opened 2020-11-25
ENH: Include plans from nabs

ZLLentz:
## Description

- In addition to the bluesky built-ins, include our plans in the standard loader too
- Remove the pcdsdaq preprocessor that was moved to nabs

## Motivation and Context

- It should be easy to find the tested plans in the interactive sessions
- closes #255 

## How Has This Been Tested?

Checked for every single nabs plan, and did a spot-check of the bluesky plans

### DIFF
--- a/hutch_python/plan_defaults.py
+++ b/hutch_python/plan_defaults.py
@@ -1,5 +1,4 @@
 from importlib import import_module
-from inspect import isgeneratorfunction
 from types import SimpleNamespace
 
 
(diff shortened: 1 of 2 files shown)


hutch_python/plan_defaults.py
```diff
@@ -18,22 +17,23 @@ def collect_plans(modules):
...
             except AttributeError:
                 # obj did not have __module__, probably a builtin
                 pass
     return SimpleNamespace(**plans)
 
 
-plans = collect_plans(['bluesky.plans'])
-plan_stubs = collect_plans(['bluesky.plan_stubs'])
+plans = collect_plans(['bluesky.plans',
+                       'nabs.plans'])
```
klauer, line 35: Worth noting: modules listed later in the list will override plan entries from the previous ones.

The ordering shown makes sense, as we'd want our custom `nabs` plans to override anything built-in from bluesky.

klauer (approved):

PR #256, opened 2020-11-25
MNT: remove the z check in happi loading

### LINKED ISSUE #247: Happi devices with negative z are ignored in load
## Expected Behavior

Load all hutch devices

## Current Behavior

Loads only `z>=0`
https://github.com/pcdshub/hutch-python/blob/493565d1e3e007f3791921e8fe4b0bf31982968a/hutch_python/happi.py#L49-L50

## Possible Solution

Load everything? I probably want to discuss this with someone to make sure there isn't a good reason for this `z>=0` stipulation.

## Context

tmo lamp was at z=-1 and didn't load in tmo3

ZLLentz:
## Description

- Remove the check for device z location in the happi loader
- Remove the undocumented configuration for start/end of beamline z

## Motivation and Context

- Frequently causes confusion: why are my devices not loading?
- Redundant behavior with the "active" tag
- closes #247 

## How Has This Been Tested?

Still passes regression tests

This is fixing a bug basically- the z behavior was not documented, so the documentation matches the code better after this change.

### DIFF
--- a/hutch_python/happi.py
+++ b/hutch_python/happi.py
@@ -2,8 +2,8 @@ import logging
 
 import happi
 import lightpath
-from lightpath.config import beamlines
 from happi.loader import load_devices
+from lightpath.config import beamlines
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


klauer (approved):
Makes sense. 👍

PR #257, opened 2020-11-25
MNT: Add calcs from pcdscalc

ZLLentz:
## Description

Include be lens, common, and diffraction calcs in the default namespace

## Motivation and Context

- Allows scientists to use them
- closes #257
- closes unrelated #258 

## How Has This Been Tested?

Interactively


klauer (commented):
I think this looks good overall for now.

Let's also look forward a bit, at a point when we have a bunch of submodules in pcdscalc.
`collect_functions`, as I understand it, will remove the module namespace hierarchy (e.g., `pcdscalc.be_lens_calcs.configure_defaults` -> `calc_namespace.configure_defaults`).
Do you think it'd be wise to avoid future clashing by adding back in the hierarchy now?

Maybe `calc_namespace.be_lens.configure_defaults` (alternatively matching `pcdscalc` naming) instead?

ZLLentz:
That's a really good point about namespace clashing.

I think it's worth fixing now before people get used to `calc.<thing>`
I think it's even desirable to sort them by source. So you can tab through `calc.<thing I want calcs for>.<calc I want>`, it also makes it easy to understand the intent of a calc function without sifting through all the docstrings.

ZLLentz:
- Avoided namespace conflicts
- Added the newest calculations from pcdscalc as well
- Tagged pcdscalc in anticipation of this PR
- Modified docstring for #258 really quick

klauer (approved):
Looks good. Bonus points for closing the unrelated issue 👍

ZLLentz:
The tests failed because the install stage used the wrong recipe... it installed pcdscalc=0.1.0 even though 0.2.0 is specified. I'm going to close and re-open to reset because the output looks strange.

klauer:
`v0.2.0` isn't up on conda-forge just yet, but should be soon (there's a bit of latency there unfortunately)

ZLLentz:
I understand that 0.2.0 isn't on conda-forge, but it used 0.2.0/pcds-tag for the first step of the travis (conda build) just fine... and then picked 0.1.0/conda-forge for the install step instead of either failing outright or picking 0.2.0/pcds-tag

klauer:
Gotcha, didn't realize 👍

ZLLentz:
The remaining failure is that if `colorama` is installed, the shelled-out `tstpython` basic tests fails because it tries to set terminal colors as if we're on a windows machine...

ZLLentz:
I ended up changing a lot of other things in the test suite before reaching this conclusion

klauer (approved):
All looks good (again) - happi fixes and reduction in deprecation warnings is nice.

klauer:
Tests now passing; merge at will if ready.

ZLLentz:
Test passing because I skipped the bad one :potato:
I think there is a bug in `colorama` but I'm having trouble understanding it and putting it into words
We are having this tests-only issue because `IPython` uses color escape sequences that are captured by `colorama`, which is activated by both `bluesky` and `ophyd` if it is importable.
I'm not sure which of our dependencies suddenly added a `colorama` dependency.

PR #267, opened 2021-04-13
ENH/REF: IterableNamespace -> HelpfulNamespace

klauer:
## Description
Useful tables show in IPython for our namespaces now, even recursively.
There's hooks for showing the tables - with some caveats for recursively showing tables - in HTML as well.

## Motivation and Context
Display useful tables!

## How Has This Been Tested?
Simple tests in the suite.

## Where Has This Been Documented?
Docstrings and release notes.

## Screenshots (if appropriate):

<details>
<summary>Recursive output</summary>

```
In [1]: all_objects
Out[1]:
This HelpfulNamespace has the following attributes available:

+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
| Attribute  |       Class        | Description                                                                                                                 |
+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
|     RE     |     RunEngine      | The Run Engine execute messages and emits Documents.                                                                        |
+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
|  archive   |    EpicsArchive    | Interactive interface to an Archive Appliance                                                                               |
+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
| beam_stats |     BeamStats      |                                                                                                                             |
+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
|    bec     | BestEffortCallback |                                                                                                                             |
+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------+
|     bp     |  HelpfulNamespace  | +-----------------------------+----------+-----------------------------------------------------------------------------+    |
|            |                    | |          Attribute          |  Class   | Description                                                                 |    |
|            |                    | +-----------------------------+----------+-----------------------------------------------------------------------------+    |
|            |                    | |        adaptive_scan        | function | Scan over one variable with adaptively tuned step size.                     |    |
|            |                    | |            count            | function | Take one or more readings from detectors.                                   |    |
|            |                    | |          daq_a2scan         | function | Two-dimensional daq scan with absolute positions.                           |    |
|            |                    | |          daq_a3scan         | function | Three-dimensional daq scan with absolute positions.                         |    |
|            |                    | |          daq_ascan          | function | One-dimensional daq scan with absolute positions.                           |    |
|            |                    | |          daq_count          | function | Take repeated DAQ runs with no motors.                                      |    |
|            |                    | |        daq_delay_scan       | function | Scan a laser delay timing motor with DAQ support.                           |    |
[...]

PR #268, opened 2021-04-13
ENH: load_cache doc annotations

ZLLentz:
## Description

- Annotate load_conf.py with docs for each item added
- Make sure these doc strings get included in the namespace table summaries
- Reorganize so that all items are added to the cache in the same way (previously, some debug utilities and misc were strapped on after the normal load process)

## Motivation and Context

- Need some top-level documentation as a companion to #267 which works great for everything that isn't too generic
- E.g. namespaces will now have  quick blurb about what exactly they are namespaces for

## How Has This Been Tested?

- Existing tests still pass

- Expanded release notes

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -1,7 +1,7 @@
 Release History
 ###############
 
-v1.9.5 (2021-??-??)
+v1.9.2 (2021-??-??)
 ===================
 
 Features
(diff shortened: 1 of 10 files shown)


klauer (approved):
Gave this a spin locally. Works pretty great, and code is clean, too.

Only one minor thing is:
`bp?` gives
image

whereas `bp` gives:
image

I think it's more than good enough for now though.

ZLLentz:
ah bizarre, I wonder what went wrong there

ZLLentz:
I see, the pretty table has a hard-coded table header while the `__doc__` is just my extra added text plus the table

klauer:
Slightly wonky implementation from me is what happened 😁 - `__repr__` (generated dynamically) is different from `__doc__` (generated at init)

ZLLentz:
Ok I'm just going to address the test failure then (checking terminal width fails if there is no terminal)

PR #276, opened 2021-06-03
ENH: set defaults for ophyd epics signal timeouts

### LINKED ISSUE #273: Use EpicsSignal.set_defaults to configure default timeouts
## Expected Behavior
Infinite timeouts are (arguably _!!_ ) a bad thing, and can potentially cause frequent "Another set() call is currently in progress" issues.

## Current Behavior
We don't set a timeout, and rely on ophyd to provide sane defaults.

## Possible Solution
Use `EpicsSignalBase.set_defaults` to configure the timeouts:
https://github.com/bluesky/ophyd/blob/c9f44e21ddd2b722a404bae4bce45d99b263c333/ophyd/signal.py#L828-L852

## Context
XCS had a very confusing timeout-related bug that took much too long to diagnose.

ZLLentz:
## Description

- Set the default ophyd epics signal write_timeout to 5.0 instead of no timeout (previously: wait forever)
- Explicitly pick the current ophyd defaults as the pcds defaults for the other parameters

## Motivation and Context

- Nothing in LCLS needs an infinitely long write timeout
- We had problems in XCS with infinite write timeouts disrupting later signal puts (the set thread blocks if a set is already in progress)
- Explicitly opt in to finite timeouts for everything else in case some ophyd default changes to infinite in the future
- closes #273 

## How Has This Been Tested?

Ran with TMO's config to verify that it was run early enough in the chain:
```
 _                   _____       _   _
| |                 |  __ \     | | | |
| |_ _ __ ___   ___ | |__) |   _| |_| |__   ___  _ __
| __| '_ ` _ \ / _ \|  ___/ | | | __| '_ \ / _ \| '_ \
| |_| | | | | | (_) | |   | |_| | |_| | | | (_) | | | |
 \__|_| |_| |_|\___/|_|    \__, |\__|_| |_|\___/|_| |_|
                            __/ |
                           |___/

INFO     Selected default hutch-python daq platform: 0
INFO     Loading debug tools...
SUCCESS  Successfully loaded debug tools in 0.00 s
INFO     Loading options...
SUCCESS  Successfully loaded options in 0.00 s
INFO     Loading configure ophyd...
SUCCESS  Successfully loaded configure ophyd in 0.00 s
INFO     Loading run engine...
SUCCESS  Successfully loaded run engine in 0.00 s
```

Checked the signal write timeouts:
```
In [5]: from ophyd.signal import EpicsSignal

In [6]: sig = EpicsSignal('stuff', name='stuff')

In [7]: sig.write_timeout
Out[7]: 5.0
```

Also did not break existing tests. Did not write new tests because I don't want reflexive test behavior where the code picks a number and the test just checks against the number, a common maintenance hazard in cases where we may want to change the settings later.

Will make it into the release notes.
Added into the load_conf docstring

### DIFF
--- a/hutch_python/ophyd_settings.py
+++ b/hutch_python/ophyd_settings.py
@@ -12,6 +12,9 @@ READ_VAR = 'HUTCH_PYTHON_READ_TIMEOUT'
 WRITE_VAR = 'HUTCH_PYTHON_WRITE_TIMEOUT'
 AUTO_VAR = 'HUTCH_PYTHON_AUTO_MONITOR'
 
+# Acceptable values to make auto_monitor True
+AUTO_TRUE = {'true', 't', 'yes', 'y', '1'}
+
 
 def setup_ophyd():
     """
@@ -37,8 +40,8 @@ def setup_ophyd():
       - write timeout (we don't want to wait forever)
     """
     EpicsSignalBase.set_defaults(
-        connection_timeout=os.environ.get(CONN_VAR, 1.0),
-        timeout=os.environ.get(READ_VAR, 2.0),
-        write_timeout=os.environ.get(WRITE_VAR, 5.0),
-        auto_monitor=os.environ.get(AUTO_VAR, False),
+        connection_timeout=float(os.environ.get(CONN_VAR, 1.0)),
+        timeout=float(os.environ.get(READ_VAR, 2.0)),
+        write_timeout=float(os.environ.get(WRITE_VAR, 5.0)),
+        auto_monitor=os.environ.get(AUTO_VAR, 'false').lower() in AUTO_TRUE,
         )
--- a/hutch_python/ophyd_settings.py
+++ b/hutch_python/ophyd_settings.py
@@ -1,0 +1,29 @@
+"""
+Module to contain ophyd-specific settings that must be applied prior to
+creating any ophyd objects.
+"""
+from ophyd.signal import EpicsSignalBase
+
+
+def setup_ophyd():
+    """
+    Call EpicsSignalBase.set_defaults with pcds defaults.
+
+    Sets the following to the current ophyd defaults:
+      - connection_timeout
+      - timeout (read timeout)
+      - auto_monitor
+
+    Overriding defaults with themselves seems redundant, but I want to
+    be insulated from changes to the upstream consensus.
+
+    Sets a new default to the following:
+      - write timeout (we don't want to wait forever)
+    """
+    EpicsSignalBase.set_defaults(
+        connection_timeout=1.0,
+        timeout=2.0,
+        write_timeout=5.0,
+        auto_monitor=False,
+        )
+
--- a/hutch_python/ophyd_settings.py
+++ b/hutch_python/ophyd_settings.py
@@ -2,12 +2,28 @@
 Module to contain ophyd-specific settings that must be applied prior to
 creating any ophyd objects.
 """
+import os
+
 from ophyd.signal import EpicsSignalBase
+
+# Environment variable definitions
+CONN_VAR = 'HUTCH_PYTHON_CONNECTION_TIMEOUT'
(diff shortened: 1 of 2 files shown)


hutch_python/ophyd_settings.py
```diff
@@ -0,0 +1,29 @@
...
+      - auto_monitor
+
+    Overriding defaults with themselves seems redundant, but I want to
+    be insulated from changes to the upstream consensus.
+
+    Sets a new default to the following:
+      - write timeout (we don't want to wait forever)
+    """
+    EpicsSignalBase.set_defaults(
+        connection_timeout=1.0,
```
klauer, line 24: Just a thought:
Would you consider making each of these an environment variable, such that we could tweak per-environment settings if this ends up being an issue on a larger scale than expected?

I'd imagine this will only affect 1 or 2 devices in a given hutch, but say for the sake of the argument it ends up affects like 50 of them. Then it starts to make more sense to just increase the environment-specific settings (or wait until hutch-python gets updated).

`HUTCH_PYTHON_CONNECTION_TIMEOUT=1.0` and whatnot

klauer (approved):
I like the general hook, and it seems good enough to me.

If you think the env variable suggestion is too complicated/unwarranted/etc, then I think this is ready to go.

hutch_python/ophyd_settings.py
ZLLentz: This is easy enough to do that it's worth doing

ZLLentz: I now recall that we've occasionally had issues (in various apps) with e.g. slow waveforms from the gateway, which can be solved by extending the read timeout. Therefore, being able to manipulate this via environment variables already has an observed use case.

hutch_python/ophyd_settings.py
```diff
@@ -0,0 +1,44 @@
...
+      - auto_monitor
+
+    Overriding defaults with themselves seems redundant, but I want to
+    be insulated from changes to the upstream consensus.
+
+    Sets a new default to the following:
+      - write timeout (we don't want to wait forever)
+    """
+    EpicsSignalBase.set_defaults(
+        connection_timeout=os.environ.get(CONN_VAR, 1.0),
```
klauer, line 40:
```suggestion
        connection_timeout=float(os.environ.get(CONN_VAR, 1.0)),
```

hutch_python/ophyd_settings.py
```diff
@@ -0,0 +1,44 @@
...
+        timeout=os.environ.get(READ_VAR, 2.0),
+        write_timeout=os.environ.get(WRITE_VAR, 5.0),
+        auto_monitor=os.environ.get(AUTO_VAR, False),
```
klauer, line 43:
```suggestion
        auto_monitor=os.environ.get(AUTO_VAR, "false").lower() in {"false", "no", "0"},
```

klauer (commented):
Why can't environment variables have a type other than `str`? Oh `/bin/sh`...

hutch_python/ophyd_settings.py
ZLLentz: ahh good catch

ZLLentz: Implemented as suggested in https://github.com/pcdshub/hutch-python/pull/276/commits/5d3c6c17b8bc2d59a66c2e17e81fdca543cc67ac

hutch_python/ophyd_settings.py
ZLLentz: alternate implementation in https://github.com/pcdshub/hutch-python/pull/276/commits/5d3c6c17b8bc2d59a66c2e17e81fdca543cc67ac

klauer:
LGTM!

ZLLentz:
One more check with actually setting all of the env vars before I hit merge

PR #277, opened 2021-06-03
DOC: add release notes for v1.10.1

ZLLentz:
Small release to maybe hit an environment tag today

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -1,6 +1,20 @@
 Release History
 ###############
 
+v1.10.1 (2021-06-03)
+===================
+
+Bugfixes
+--------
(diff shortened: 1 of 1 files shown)


klauer (approved):
Building the general defaults around the functionality of the minority always seemed the wrong approach to me.

That said, we do have some signals which behave this way - at least to my recollection (corrections welcome):
1. AreaDetector acquire blocks until stopped
2. Some positioners use put completion to notify clients of motion complete status (see `PVPositionerPC` - though we might just ignore this in practice
3. caproto IOCs use this functionality by default - put complete callback when your `pvproperty` `putter` returns (so: solid attenuators for now, pretty much)

ZLLentz:
Thanks for the correction- I guess I just assume that things I don't know about do not exist. Whoops.

PR #279, opened 2021-06-03
WIP: Selective verbose logging of ophyd objects

klauer:


ZLLentz:
I like the way this is set up a lot

docs/source/releases.rst
```diff
@@ -1,6 +1,25 @@
 Release History
 ###############
 
+v1.?.? (2021-??-??)
```
ZLLentz, line 4: Are you interested in just setting this to v1.11.0 and tagging early? I don't remember any other immediate-priority updates planned in this repo.

```suggestion
v1.11.0 (2021-06-04)
```

klauer: I don't mind, either way 👍

ZLLentz (approved):
Looks good to me. You made a lot of seemingly small but definitely important changes in addition to the logging features that I appreciated. Adding the logs namespace was also a good call. I wish I had suggestions for additions/changes but I do not.

docs/source/releases.rst
ZLLentz: In that case I suggest we do it. Version numbers are cheap and it saves us some work later.

klauer:
Thanks, @ZLLentz!

We may want to do a quick dev installation test for some given hutch, as I've only tested this locally and through the test suite. While I don't expect it will cause any issues, it's better safe than sorry.

ZLLentz:
Oh, sure. It's not hard to set up a quick test like that.

hutch_python/log_setup.py
```diff
@@ -1,74 +1,366 @@
...
     # Configure centralized PCDS logging:
     pcdsutils.log.configure_pcds_logging()
+
+    # This ensures that centralized logging messages do not make it to the
+    # user or other log files.
     central_logger.propagate = False
 
     logging.config.dictConfig(config)
-    noisy_loggers = ['parso', 'pyPDB.dbd.yacc', 'ophyd', 'bluesky']
+    noisy_loggers = ['ophyd.event_dispatcher', 'parso', 'pyPDB.dbd.yacc', 'bluesky']
```
ZLLentz, line 155:
```suggestion
    noisy_loggers = ['ophyd.event_dispatcher', 'parso',
                     'pyPDB.dbd.yacc', 'bluesky']
```

PR #280, opened 2021-06-15
ENH: fix issues related to matplotlib setup in headless mode

ZLLentz:
## Description

- Remove `matplotlib.use()`
- Use the environment variable as the principal way to determine the matplotlib backend
- Disable the best effort callback plotting if we don't have a qt5 backend
- Set up the hutch-python ipython session to match the same backend determined in the load process

## Motivation and Context

- `matplotlib.use()` can fail if we run in a headless mode
- the environment variable falls back to `agg`, which works in headless mode
- bec fails if we don't use qt5
- ipython needs special setup for matplotlib

## How Has This Been Tested?

- interactively
- hopefully tests still pass

- will be in the release notes

### DIFF
--- a/hutch_python/mpl_config.py
+++ b/hutch_python/mpl_config.py
@@ -2,5 +2,7 @@
 # If no environment variable set, try qt5 (best for hutch-python)
 import os
 os.environ.setdefault('MPLBACKEND', 'Qt5Agg')
-# Force matplotlib to pick a backend
+# Matplotlib will choose an appropriate backend based on the environment variable
+# settings and its ability to communicate with an X server. Our preference, as above,
+# is to use the Qt5 backend when available.
 import matplotlib.pyplot
--- a/hutch_python/mpl_config.py
+++ b/hutch_python/mpl_config.py
@@ -1,4 +1,3 @@
-# noqa
 # If no environment variable set, try qt5 (best for hutch-python)
 import os
 os.environ.setdefault('MPLBACKEND', 'Qt5Agg')
--- a/hutch_python/mpl_config.py
+++ b/hutch_python/mpl_config.py
@@ -1,7 +1,7 @@
 # If no environment variable set, try qt5 (best for hutch-python)
 import os
 os.environ.setdefault('MPLBACKEND', 'Qt5Agg')
-# Matplotlib will choose an appropriate backend based on the environment variable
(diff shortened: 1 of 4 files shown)


hutch_python/mpl_config.py
```diff
@@ -1,2 +1,6 @@
-import matplotlib
-matplotlib.use('Qt5Agg')
+# noqa
```
klauer, line 1: Does this work on a single line? Interesting...

hutch_python/mpl_config.py
```diff
@@ -1,2 +1,6 @@
...
+# If no environment variable set, try qt5 (best for hutch-python)
+import os
+os.environ.setdefault('MPLBACKEND', 'Qt5Agg')
+# Force matplotlib to pick a backend
```
klauer, line 5:
```suggestion
# Matplotlib will choose an appropriate backend based on the environment variable
# settings and its ability to communicate with an X server. Our preference, as above,
# is to use the Qt5 backend when available.
```

klauer (approved):
Interactive testing is really key here (and I know you've done due diligence). Just added some comment notes for clarification, for your consideration.

hutch_python/mpl_config.py
klauer: Argh, unfortunately it does not. That's what's causing the CI failures:
```
hutch_python/mpl_config.py:6:1: E402 module level import not at top of file
hutch_python/mpl_config.py:6:1: F401 'matplotlib.pyplot' imported but unused
```
Instead, I think, this should be added as an exception in `/.flake8`, perhaps:

```
[flake8]
exclude = .git,__pycache__,build,dist,versioneer.py,hutch_python/mpl_config.py
```

hutch_python/mpl_config.py
```diff
@@ -1,2 +1,6 @@
...
+import matplotlib.pyplot
```
klauer, line 6:
```suggestion
import matplotlib.pyplot  # noqa: F401 E402
```

hutch_python/mpl_config.py
```diff
@@ -1,2 +1,8 @@
-import matplotlib
-matplotlib.use('Qt5Agg')
+# noqa
```
ZLLentz, line 1:
```suggestion
```

PR #288, opened 2021-08-09
DOC: v1.11.2, MNT: optional deps

ZLLentz:
## Description

- release notes
- make elog and lightpath optional to match the updated requirements.txt

## Motivation and Context

- other modules are being tagged, let's tag here too
- let's clean up the optional deps if we are tagging, should be able to run this from pip for dev

## How Has This Been Tested?

- interactively only to make sure the user warning messages looked correct

- release notes

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -1,6 +1,22 @@
 Release History
 ###############
 
+v1.11.2 (2021-08-09)
+====================
+
+Fixes and Maintenance
+---------------------
+- Fix order of message logging in the IPython input logger. Previously, the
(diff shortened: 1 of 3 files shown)


klauer (approved):
Looks good to me. Glad to see elog made optional.

I think lightpath is a good optional dependency as well, even if we do have it on pypi. https://pypi.org/project/lightpath/

ZLLentz:
I was going off the "broken on pypi" comment in the requirements.txt file- didn't notice that it was in the requirements list above still!

klauer:
My memory is faulty - I even made that "broken on PyPI" comment... I think. 😁

PR #289, opened 2021-09-21
ENH: LCLS2 DAQ Instantiation, MNT/CI: Various Fixes, Compat

ZLLentz:
## Description

Create the daq object automatically for LCLS2 hutches. This relies on the support provided by Chris Ford in psdaq.control and will require some additions to the hutch python configuration file to specify things like the drp host and platform.

Other changes:
- Adjust the CI allow_failures for the new name of the PIP build, and require all the conda builds to pass
- Sort the requirements and pin latest for build speed
- Extend the documentation to cover our new configuration options
- make psdm_qs_cli an optional dependency because it doesn't install properly with pypi
- remove lightpath and psdm_qs_cli from requirements.txt because they don't install properly with pypi
- make elog an optional dependency in the tests (was already optional in the module)
- make lightpath an optional dependency in the tests (was already optional in the module)
- mark all the tests that fail on windows with skip-ifs

## Motivation and Context

Currently we're creating the daq in a weird different way in each of TMO/RIX/UED. Instead, do it in a central place, and make it look as similar to what we're doing with the LCLS1 DAQ as possible: that is, the daq should always be instantiated, and always in the same way, and the user should be able to grab it and use it in data collection right away with no additional setup.

The other changes were geared toward allowing me to run/test offline on my windows laptop and to allow the pypi test to run quickly instead of taking 30 minutes to fail.

## How Has This Been Tested?

Added relevant testing

Added config docs


klauer:
Would a variant like `lcls1-sim` for pcdsdaq's `set_sim_mode(True)` be worthwhile? Maybe not enough to centralize it for common usage.

ZLLentz:
I think so, and it should be too hard to do.

ZLLentz:
Ok, actually this fails to start the CI because of merge conflicts in my last commit, I must have not started from current master.

ZLLentz:
Hmm, it's a bit frustrating, but I might have to nix the pypi build/test on this repo. It takes longer to build the pypi environment than the conda one in this case... There's got to be some way to speed this up (or maybe there's a conflict still?)

ZLLentz:
Oh boy finally green, I need to update the PR description properly now and swindle someone into letting me test on the real DAQ

ZLLentz:
Also needs +1 issue for the pip test failures

ZLLentz:
Remaining todo here is to add a test that configures for the lcls2 daq and checks that an object of the correct type is created. This might require some monkeypatching to avoid real connections and timeouts.

ZLLentz:
Planning to test this in TMO today

docs/source/yaml_files.rst
```diff
@@ -85,6 +86,35 @@ to load an experiment other than the active experiment; handy for debugging.
...
 particular experiment.
 
+daq_type
+--------
+
+The ``daq_type`` key is optional. If omitted, the default value is 'lcls1'
+for backwards compatibility with existing hutch python setups.
+This key expects a string with one of four valid values:
+'lcls1', 'lcls1-sim', 'lcls2', or 'nodaq', to pick between creating an
+LCLS1-style daq, a simulated LCLS1-styel daq, an LCLS2-style daq,
```
klauer, line 96:
```suggestion
LCLS1-style daq, a simulated LCLS1-style daq, an LCLS2-style daq,
```

docs/source/yaml_files.rst
```diff
@@ -85,6 +86,35 @@ to load an experiment other than the active experiment; handy for debugging.
...
+daq_platform
+------------
+
+A dictionary description of which daq platform to use. This is used to
+determine whether to use the primary or secondary elog in hutches with
+two daqs, and is used to set up the lcls2 daq_type. If omitted entirely,
+platform 0 and the primary elog will be used.
+This dictionary has a required key, "default" that points to an integer
+that is the normal platform to use, associated with the primary
+experiment. Additional keys are interpretted as hostnames to use
```
klauer, line 115:
```suggestion
experiment. Additional keys are interpreted as hostnames to use
```

hutch_python/load_conf.py
```diff
@@ -128,17 +130,20 @@ def load_conf(conf, hutch_dir=None):
 
     Parameters
     ----------
-    conf: ``dict``
+    conf : ``dict``
```
klauer, line 133: May as well fix up the types here while you're touching the line:
```suggestion
    conf : dict
```

hutch_python/load_conf.py
```diff
@@ -206,9 +211,33 @@ def load_conf(conf, hutch_dir=None):
                          'load objects from questionnaire or experiment '
                          'file.'))
 
+    try:
+        # Configure whether we use the LCLS-I or LCLS-II daq
+        daq_type = conf['daq_type']
+        if daq_type in ('lcls1', 'lcls1-sim', 'lcls2', 'nodaq'):
```
klauer, line 217: Nitpick - should move this out of the try/except as it won't be raising `KeyError` (or in `else:`)

hutch_python/load_conf.py
```diff
@@ -293,8 +322,40 @@ def load_conf(conf, hutch_dir=None):
 
     # Daq
     with safe_load('daq'):
-        cache(daq=Daq(RE=RE, hutch_name=hutch))
-        cache.doc(daq='LCLS1 DAQ interface object.')
+        if daq_type.startswith('lcls1'):
+            if daq_type == 'lcls1-sim':
```
klauer, line 326:
```suggestion
            if daq_type == 'lcls1-sim' or (args and args.sim):
```
Maybe?

hutch_python/tests/conftest.py
```diff
@@ -158,10 +163,50 @@ def fake_curexp_script():
...
+    def send_json(self, *args, **kwargs):
+        pass
+
+    def recv_json(self, *args, **kwargs):
+        return {}
+
+
+@pytest.fixture(scope='function')
+def dummy_zmq_lcls2(monkeypatch):
+    # monkeypatch such that the lcls2 daq doesn't actually connect to things
```
klauer, line 211: 😁

hutch_python/tests/test_cli.py
```diff
@@ -24,6 +25,7 @@ def no_op(*args, **kwargs):
     monkeypatch.setattr(hutch_python.cli, 'start_ipython', no_op)
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Does not run on Windows")
```
klauer, line 28: Just a suggestion, but maybe:
conftest.py -> `skip_if_win32 = pytest.mark.skipif(sys.platform == "win32", reason="Does not run on Windows")`

Then `@skip_if_win32`

klauer (approved):
Looks good to me - just some minor, optional suggestions.

I like the direction you're going with cross-platform compatibility checks in the test suite, optional deps, and so on. 👍

hutch_python/load_conf.py
ZLLentz: Forgot about this one, it looked correct when I read through it

hutch_python/load_conf.py
ZLLentz: Hmm, seems good- maybe combined with removing the other place where the code sets the tim

ZLLentz: Cleans it up to avoid setting the sim in two places

hutch_python/load_conf.py
ZLLentz: I think this is a difference in intent rather than a straight error.

The intent of the try/except is just to catch a missing key to provide the default. I don't want to fall back to the default if the user passes an invalid key, I want to make red text in the startup and not load the daq in the session. Config errors should be loud I think. Maybe this one should be even louder?

hutch_python/tests/test_cli.py
ZLLentz: That's a good idea

hutch_python/load_conf.py
klauer: My terse comment wasn't as clear as I intended - I meant "keep the `try` section small and only relevant to what could be raised; the first line raising `KeyError` is the only relevant one"

ZLLentz: Ah, I understand now. I'll move these into the often-forgotten try/except/else

ZLLentz:
As a bonus, moving the skipifs to conftest now means that we can test for optional dependencies all in one place, rather than doing import try/excepts in every sub test module!

I've tested with the real DAQ in TMO and everything seemed to work

ZLLentz:
Ok, it seems like tmo diverged from rix and this really only works with rix's SOP. tmo runs in a no-bluesky mode sometimes so in different sessions their daq object can be a different object. Gross.

There will need to be a follow-up at some point- I'm not sure what path I'm taking. I do have some desire to take the `pcdsdaq` interface and use it almost 1-to-1 in the new hutches with some compatibility layer- it might be the easiest thing to do.

ZLLentz:
Some more context, the no-bluesky mode uses a different object because the bluesky object doesn't allow for manual stop/starts outside of scans

Issue #297, opened 2021-10-15
Add ability to ignore (specific?) upstream devices in load


ZLLentz:
## Expected Behavior

Hutch user should get exactly which objects they want and none (or few) extra

## Current Behavior

Hutch user currently gets all hutch devices plus all upstream devices based on the lightpath

## Possible Solution

- config/cli option (`--no-upstream`? `--no-lightpath`?)
- more precise config control (`--beamlines xpp lfe`, `--ignore xpp`)

## Steps to Reproduce (for bugs)

1. Open a session in e.g. mec who is very far downstream

## Context

Does MEC need to see LODCM errors? I don't think they care.

ZryletTC:
This makes sense and is somewhat similar to another thing I think is desired by xpp, not loading camViewer when desired. I was playing around with adding an option to the conf.yaml file. What do you think of putting these options in there? I would also vote for more fine control than just a `--no-upstream` or similar option.

klauer:
What if you could specify sets of `happi search` queries you wanted to include?

PR #298, opened 2021-10-19
ENH: record and optionally filter noisy ophyd loggers

### LINKED ISSUE #296: Add ability to blacklist certain ophyd objects from logging
Some ophyd object logs can be spammy. 

Until we can fix the source of the issue, provide tools to filter them out.

https://github.com/pcdshub/hutch-python/blob/6d9b47a926084fff12fce27392a6dbaa88300e17/hutch_python/log_setup.py#L196

klauer:
## Description
Functionality proposal, at least.

* Add in the possibility of automatically filtering out ophyd loggers based on the number of messages they emit
* Done on a per-object basis, in periods of 1s, 10s, and 60s.
* Escape-hatch on a per-object basis as well

## Motivation and Context
* Our ophyd object definitions are imperfect
* We can and will fix noisy log messages, but they can be crippling for end-users

I think this proposal:
* Gives users an additional hook to catch new things that arise and still finish their experiments
* Gives us a bit of room to fix our imperfect devices more at our leisure
* Could be done in addition to other things

Bonus: closes #296 

Related: #297 

## How Has This Been Tested?
Interactively, for now

Pending agreement

## Screenshots (if appropriate):

```python
In [3]: for i in range(200): sim.fast_motor2.log.warning("test")
...
INFO     _count_update - Hushing noisy logger 'fast_motor2'. If you see this often, please consider reporting it to your POC or #pcds-help.  If this functionality is undesirable, adjust `noisy_logger_whitelist`.
```
or
```python
In [1]: flt1.blacklist.add("fast_motor1")

In [2]: sim.fast_motor1.log.warning("test")

In [3]: flt1.blacklist.clear()

In [4]: sim.fast_motor1.log.warning("test")
WARNING  <module> fast_motor1 test
```

### DIFF
--- a/hutch_python/log_setup.py
+++ b/hutch_python/log_setup.py
@@ -239,19 +239,20 @@ class ObjectFilter(logging.Filter):
         If a single ophyd object logs over ``noisy_threshold_60s`` log messages
         in 60 seconds, consider it a noisy logger and silence it.
 
-    logger_whitelist : list of str, optional
-        Logger names that are not subject to the thresholds above.
+    whitelist : list of str, optional
+        Logger or object names that are not subject to the thresholds above.
 
-    logger_blacklist : list of str, optional
-        Logger names that should always be filtered out.
+    blacklist : list of str, optional
+        Logger or object names that should always be filtered out.
 
     Attributes
     ----------
-    logger_blacklist : list of str
+    blacklist : list of str
         Logger names that should always be filtered out.
 
-    logger_whitelist : list of str
-        List of noisy loggers that are exempt from the noise thresholds.
+    whitelist : list of str
+        List of noisy loggers or object names that are exempt from the noise
+        thresholds.
 
     noisy_loggers: set of str
         Loggers marked as noisy and to be filtered out, unless in the
@@ -269,9 +270,9 @@ class ObjectFilter(logging.Filter):
     noisy_threshold_1s: int
     noisy_threshold_10s: int
     noisy_threshold_60s: int
-    logger_whitelist: list[str]
+    whitelist: list[str]
     noisy_loggers: set[str]
-    logger_blacklist: set[str]
+    blacklist: list[str]
 
     def __init__(
         self,
@@ -281,8 +282,8 @@ class ObjectFilter(logging.Filter):
         noisy_threshold_1s: int = 20,
         noisy_threshold_10s: int = 50,
         noisy_threshold_60s: int = 100,
-        logger_whitelist: Optional[list[str]] = None,
-        logger_blacklist: Optional[list[str]] = None,
+        whitelist: Optional[list[str]] = None,
+        blacklist: Optional[list[str]] = None,
         allow_other_messages: bool = True
     ):
         self._objects = frozenset(objects)
@@ -295,8 +296,8 @@ class ObjectFilter(logging.Filter):
         self.noisy_threshold_1s = int(noisy_threshold_1s)
         self.noisy_threshold_10s = int(noisy_threshold_10s)
         self.noisy_threshold_60s = int(noisy_threshold_10s)
-        self.logger_whitelist = list(logger_whitelist or [])
-        self.logger_blacklist = list(logger_blacklist or [])
+        self.whitelist = list(whitelist or [])
+        self.blacklist = list(blacklist or [])
         self.noisy_loggers = set()
 
         self._running = True
@@ -329,7 +330,7 @@ class ObjectFilter(logging.Filter):
         )
 
         for noisy_logger in sorted(noisy_loggers):
-            if noisy_logger in self.logger_whitelist:
+            if noisy_logger in self.whitelist:
                 continue
             if noisy_logger in self.noisy_loggers:
                 continue
@@ -338,7 +339,7 @@ class ObjectFilter(logging.Filter):
                 "Hushing noisy logger %r. If you see this often, please "
                 "consider reporting it to your POC or #pcds-help.  If this "
                 "functionality is undesirable, adjust the thresholds or set "
-                "`logger_whitelist`.",
+                "`whitelist`.",
                 noisy_logger
             )
             self.noisy_loggers.add(noisy_logger)
@@ -371,7 +372,7 @@ class ObjectFilter(logging.Filter):
             f"noisy_threshold_1s={self.noisy_threshold_1s}, "
             f"noisy_threshold_10s={self.noisy_threshold_10s}, "
             f"noisy_threshold_60s={self.noisy_threshold_60s}, "
-            f"logger_whitelist={self.logger_whitelist}, "
+            f"whitelist={self.whitelist}, "
             f"noisy_loggers={self.noisy_loggers}"
             f")"
         )
@@ -413,8 +414,8 @@ class ObjectFilter(logging.Filter):
 
             Loggers
             -------
-            * Block these loggers entirely: {self.logger_blacklist}
-            * Allow these noisy loggers: {self.logger_whitelist}
+            * Block these loggers/objects entirely: {self.blacklist}
+            * Allow these noisy loggers/objects: {self.whitelist}
             * Hush loggers with {self.noisy_threshold_1s} messages in 1s
             * Hush loggers with {self.noisy_threshold_10s} messages in 10s
             * Hush loggers with {self.noisy_threshold_60s} messages in 60s
@@ -423,12 +424,14 @@ class ObjectFilter(logging.Filter):
 
             Usage
             -----
-            * To allow a noisy logger through:
-                >>> filter.logger_whitelist.append('logger_name')
-            * To always filter out a logger:
-                >>> filter.logger_blacklist.append('logger_name')
-            * To whitelist specific ophyd objects:
-                >>> filter.objects = [obj1, obj2]
+            * To allow a noisy logger or ophyd device through:
+                >>> logs.filter.whitelist.append('logger_name')
+            * To always filter out a logger or ophyd device:
+                >>> logs.filter.blacklist.append('logger_name')
+            * To focus on specific ophyd objects:
+                >>> logs.log_objects(obj1, obj2)
+            * To stop
+                >>> logs.log_objects_off()
             """.rstrip()
         )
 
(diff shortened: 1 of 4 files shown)


ZLLentz (approved):
I like this approach a lot

ZLLentz:
Probably need to sit down and think about how many messages are enough to be annoying
It's also worth considering if this should be "number of messages" or "number of duplicate messages"

klauer:
Good points. I like the duplicate check idea but also worry that they would have to be checked for a similarity threshold, in case they have things like timestamps baked in and whatnot - making it even more complicated.

klauer:
I'd like to add a top-level `log` object for controlling log configuration.
`log.set_thresholds()`
`log.block(obj)`
`log.reset()`

And things like that. Having these settings buried in a log handler is inconvenient.

ZLLentz:
I think a top-level `log` object that gets loaded automatically is a good call. It gives us a great entry point for a user that wants to adjust how a specific device behaves.

klauer:
Also discussed today was: 
* What if background exceptions were always captured, bundled, and shown on request
* What if we could modify the prompt to show that information? (or would that be terrible?)

ZLLentz:
From googling, it looks like it's possible to modify the prompt. We need to be careful for the user's sake not to make it too long because that will harm the usability, but in principle it can be done.

klauer:
Fancy repr now shows:
```
In [1]: logs.filter
Out[1]:
Objects
-------
* Allow log messages at level: DEBUG
* Show messages from objects: []

Loggers
-------
* Block these loggers entirely: []
* Allow these noisy loggers: []
* Hush loggers with 20 messages in 1s
* Hush loggers with 50 messages in 10s
* Hush loggers with 50 messages in 60s
* These loggers have been identified as noisy:
[]

Usage
-----
* To allow a noisy logger through:
    >>> filter.logger_whitelist.append('logger_name')
* To always filter out a logger:
    >>> filter.logger_blacklist.append('logger_name')
* To whitelist specific ophyd objects:
    >>> filter.objects = [obj1, obj2]
```

hutch_python/log_setup.py
```diff
@@ -269,16 +466,17 @@ def whitelist_all_level(self) -> str:
...
     @property
-    def object_names(self):
+    def object_names(self) -> set[str]:
         """The names of all contained objects."""
         return set(obj.name for obj in self.objects)
 
-    def filter(self, record):
-        if not hasattr(record, 'ophyd_object_name'):
+    def filter(self, record: logging.LogRecord) -> bool:
+        name: Optional[str] = getattr(record, "ophyd_object_name", None)
```
ZLLentz, line 478: For some reason I never considered adding type annotations to variables inside function scopes. I might start doing this.

ZLLentz:
Everything here looks good to me- didn't notice anything off. I'll look again after the unit test.

hutch_python/tests/test_log_setup.py
```diff
@@ -197,3 +198,55 @@ def find_filters():
...
+    object_filter._count_update()
+    assert "Hushing noisy logger" in caplog.text
+    assert "hutch_python.tests.test_log_setup" in caplog.text
+
+    assert logger.name in object_filter.noisy_loggers
+
+
+def test_log_noisy_whitelist(caplog, object_filter: log_setup.ObjectFilter):
+    object_filter.noisy_threshold_1s = 5
+    object_filter.whitelist = ["hutch_python.tests.test_log_setup"]
```
ZLLentz, line 239:
```suggestion
    object_filter.whitelist = [logger.name]
```

hutch_python/logging.yml
```diff
@@ -33,8 +33,22 @@ handlers:
...
     level: DEBUG
+    # Allow these loggers through, even if noisy:
+    whitelist: ["pcdsdevices.interface", "hutch_python.utils"]
+    # Do not show these loggers (or ophyd device names):
+    blacklist: []
+    # Mark loggers / ophyd objects as "noisy" if they reach message levels
+    # above the given thresholds for the listed duration:
+    noisy_threshold_1s: 20
+    noisy_threshold_10s: 50
+    noisy_threshold_60s: 100
```
ZLLentz, line 45: I've been sitting here for a bit thinking about thresholds.

My first instinct is that we want these to be lower, but then I'm not quite sure. This includes debug logs, right? So I don't want it to trip too soon either.

Maybe these are already about right.

hutch_python/tests/test_log_setup.py
```diff
@@ -197,3 +198,55 @@ def find_filters():
...
+
+    for i in range(10):
+        logger.warning("Warning")
+
+    assert object_filter.name_to_log_count_1s[logger.name] == 10
+
+    caplog.clear()
+    object_filter._count_update()
+    assert "Hushing noisy logger" in caplog.text
+    assert "hutch_python.tests.test_log_setup" in caplog.text
```
ZLLentz, line 232: Another one of these, maybe
```suggestion
    assert logger.name in caplog.text
```

hutch_python/tests/test_log_setup.py
```diff
@@ -197,3 +198,55 @@ def find_filters():
...
+
+
+@pytest.fixture(scope="function")
+def object_filter(
+    monkeypatch, log_queue: queue.Queue
+) -> log_setup.ObjectFilter:
+    def no_op(*args, **kwargs):
+        ...
+
+    monkeypatch.setattr(log_setup.ObjectFilter, "_count_update_thread", no_op)
```
ZLLentz, line 210: This was a really good solution to taking the timer out of the equation for the test suite. You've essentially taken control of the clock and can advance it yourself whenever you want.

klauer: My first inclination was to add a `start` kwarg to ObjectFilter, further complicating it. I think changing it to this was a win all-around 👍

hutch_python/log_setup.py
```diff
@@ -217,36 +226,227 @@ class ObjectFilter(logging.Filter):
...
+        """Thread for checking the per-logger statistics."""
+        while self._running:
+            self._count_update()
+            time.sleep(1.0)
+
+    def _count_update(self) -> None:
+        """Log count update - every second."""
+        noisy_loggers = set(
+            name
+            for name, count in tuple(self.name_to_log_count_1s.items())
```
ZLLentz, line 320: is the `tuple` here just to eliminate cross-thread dictionary edits while iterating?

hutch_python/logging.yml
klauer: I'm definitely not certain about these values at all. It's possible this indicates the proposal is using a flawed metric, but I can't think of a real alternative.

It only includes the debug level _if_ debug logs are shown:

```
            should_show = (
                record.levelno >= self._whitelist_all_levelno or
                name in self.object_names
            ) and name not in self.blacklist
```
A bit of confusing logic, for sure...

ZLLentz: if it's only console-visible logs then that makes it easier to think about, probably should be lower, but it also might mean that our logfiles still get flooded with debug messages

hutch_python/log_setup.py
klauer: Yes, loggers recording messages in other threads could mutate the dictionary. Alternative would be a `threading.Lock` wrapping around these, but it's a bit costlier.

hutch_python/logging.yml
ZLLentz: The other thing is that the annoyance threshold for 1-line messages and 10-line tracebacks is a bit different

klauer: Very true. Should "log message count" be inclusive of "number of lines in the message"? That seems appropriate...

ZLLentz: Maybe? But now we get really complicated because a single huge traceback would kill the log. But maybe it should?

klauer:
The approach in pcdshub/pcdsutils#39 is superior in many ways. What could we gain from reworking this on top of that, and would it be worth our time now to do it instead of later?

ZLLentz:
I'm not sure there's much reworking to be done here. The desired behavior from this PR is a bit different than the one in pcdsutils. I see them as complimentary instead of contradictory.

In pcdsutils, I took the two most common sources of terminal spam linked to real problems and created a mini-framework to demote the log levels of those spammy elements. The demotion rather than deletion is so that we still see the messages in the logfile, since they represent real problems. This is a tradeoff with disk space and log readability of course.

Here, you've created a more generically applicable filter: any object that is being rude to the user in ways we did not (could not) expect will have its future log messages vetoed. 

The only optional things I would slightly consider (and it's not clear these all are good ideas) are:
- Maintaining resettable global counters for vetoed log messages
- Counting duplicate messages instead of all messages
- Applying to unseen messages as well to avoid log file bloat

ZLLentz:
The only other thought I have immediately is that this filter might be broadly applicable even outside the scope of "ophyd objects"

klauer:
>     Maintaining resettable global counters for vetoed log messages

-> If a message is filtered out, record `count[logger_name] += 1`?

>    Counting duplicate messages instead of all messages

I think subsequent duplicate log messages may actually be bugs, unless they are the result of an inner loop or something. I have a feeling log messages may be more likely to be 99.9% similar but not identical and result in spam (contrived example: "Error: position 1.234555 out of bounds", "Error: position 1.234556 out of bounds", ...)

>     Applying to unseen messages as well to avoid log file bloat

I think I agree in principle, but in practice not somuch. Verbose log files are probably a good thing overall, and it's probably not trivial to find the appropriate veto level to get "just the right amount of log messages".

> The only other thought I have immediately is that this filter might be broadly applicable even outside the scope of "ophyd objects"

I think that's true, and in fact it's already (somewhat regrettably) putting both regular log messages and ophyd object log messages down the same code path. Trying to summarize the effects of the filter as-is would be something like: "filter out noisy log messages coming from `logger_name`. If `ophyd_object_name` is available consider that a more specific version of `logger_name` and use that instead."

ZLLentz:
Ah, okay, that all makes sense. To be clear, I think this already does what we want as-is.

>  log messages may be more likely to be 99.9% similar but not identical and result in spam (contrived example: "Error: position 1.234555 out of bounds", "Error: position 1.234556 out of bounds", ...)

This is mitigated a bit by the log record containing both the template message and the resulting templated message, so if we write the above log message using the % syntax that upsets me we can mark duplicates on "Error: position %.6f out of bounds".

klauer:
> This is mitigated a bit by the log record containing both the template message and the resulting templated message, so if we write the above log message using the % syntax that upsets me we can mark duplicates on "Error: position %.6f out of bounds".

Unless the code uses those format strings a bit too generically:
```python
logger.warning("%s", object1)
logger.warning("%s", entirely_different_object2)
```
but point taken! 😁

ZLLentz:
In light of:
>  putting both regular log messages and ophyd object log messages down the same code path

Then the only meaningful refactoring path I see is separating the noisy logger identification logic from what is nominally called the `ObjectFilter`, but this is mostly a semantic thing.

klauer:
I like the idea of keeping metrics on the noisy loggers, rather than just adding them to the list. I'll add that in before we move on with this.

klauer:
New help text:
```
In [6]: logs.filter
Out[6]:
Objects
-------
* Allow log messages at level: DEBUG
* Show messages from objects: []

Loggers
-------
* Block these loggers/objects entirely: []
* Allow these noisy loggers/objects: ['pcdsdevices.interface', 'hutch_python.utils']
* Hush loggers with 20 messages in 1s
* Hush loggers with 50 messages in 10s
* Hush loggers with 50 messages in 60s
* These loggers have been identified as noisy:
  'fast_motor1': 80 messages
  'fast_motor2': 40 messages

Usage
-----
* To allow a noisy logger or ophyd device through:
    >>> logs.filter.whitelist.append('logger_name')
* To always filter out a logger or ophyd device:
    >>> logs.filter.blacklist.append('logger_name')
* To focus on specific ophyd objects:
    >>> logs.log_objects(obj1, obj2)
* To stop
    >>> logs.log_objects_off()
```

ZLLentz:
I think this all looks great. I would like to merge this, and then put a PR on top of it to add in my recent pcdsutils things, and then see how many other repo PRs we can push through before a big tag. Any objections?

klauer:
Sounds good to me. 👍

Issue #307, opened 2022-01-11
Adjust or rethink ObjectFilter configuration


klauer:
## Expected Behavior
* Minimal intrusiveness from the filter that should be _reducing_ log spam

## Current Behavior
* Excessive debug log messages can easily trigger the `ObjectFilter` configured for the file handler
  - The user is then notified that the logger is hushed (but continues to see messages from the logger in the console)
  - The log file no longer contains the full debug log message stream

## Possible Solution
Easy options:
1. Configure the object filters differently in `logging.yml` instead of sharing the settings
   - The console one with thresholds that the user might care to tweak
   - The file one with thresholds disabled (or set appropriately high)
2. Disable the log message filtering altogether
   - I have a feeling Zach's warning/exception filter will satisfy the most common complaints

## Steps to Reproduce (for bugs)
Load up xcs Python on pcds-5.1.1

## Context
See `/reg/g/pcds/pyps/apps/hutch-python/xcs/logs/2022_01/xcsopr_11_12h32m51s.log` per @ZryletTC 

## Your Environment
pcds-5.1.1

PR #308, opened 2022-01-12
ENH: threshold disable; adjust object filter settings for file

klauer:
## Description
* Add ability to disable log-hushing filter thresholds (threshold count <= 0)
* Configures "debug" (file) handler's object filter separately from the console one
* Debug object filter disables logger silencing (-> file gets full set of messages regardless of the console)

## Motivation and Context
Addresses #307 but may not close it - we need some real testing at the beamline.

## How Has This Been Tested?
Locally

Docstrings and config files.


ZLLentz (approved):
I think this is a good solution

PR #318, opened 2022-02-07
HACK: show off a potential rewriter

klauer:
Not to be taken seriously, really - just a proof-of-concept we can wipe away when we're ready. 

Based on a private discussion about #316


klauer:
Well, I had fun

ZLLentz:
I thought it was neat

PR #324, opened 2022-03-09
ENH: Changes default Ctrl+C behavior to abort

tangkong:
## Description
Makes a new Sigint handler that aborts the RunEngine rather than requesting a pause. 
Registers this as a context_manager at RunEngine initialization
Fancy threaded tests to actually send this signal within pytest

## Motivation and Context
Various beamline scientists requested Ctrl+C abort the current run and ready the RunEngine by default, rather than pausing.  

[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-525)

There are other signals available to Unix operating systems, but I'm still looking into how to activate those.

## How Has This Been Tested?
Interactively, tests pass locally

This PR.  

## Pre-merge checklist

### DIFF
--- a/hutch_python/tests/test_utils.py
+++ b/hutch_python/tests/test_utils.py
@@ -143,28 +143,23 @@ def RE_abort():
 
 
 def test_abort_RE(RE_abort):
+    # get pid so we can send SIGINT to it specifically
+    pid = os.getpid()
+
     def wait_plan():
         # arbitrarily long wait time
         yield from bps.sleep(10)
 
-    # get pid so we can send SIGINT to it specifically
-    pid = os.getpid()
-
-    def sigint_signal():
-        time.sleep(2)
+    def sim_kill():
+        time.sleep(0.05)
         os.kill(pid, SIGINT)
 
     # send sigint in a different thread
-    thread = threading.Thread(target=sigint_signal)
-    thread.daemon = True
-    thread.start()
+    timer = threading.Timer(0.5, sim_kill)
+    timer.start()
 
-    try:
+    with pytest.raises(RunEngineInterrupted):
         RE_abort(wait_plan())
-    except RunEngineInterrupted:
-        # we expect to interrupt the RunEngine
-        # other exceptions should break the test
-        pass
 
     assert RE_abort.state == 'idle'
-    assert RE_abort._exit_status == 'abort'
+    assert RE_abort._exit_status == 'success'
--- a/hutch_python/tests/test_utils.py
+++ b/hutch_python/tests/test_utils.py
@@ -146,6 +146,7 @@ def RE_abort():
         RE.halt()
 
 
+@skip_if_win32_generic
 def test_sigint_RE(RE_abort):
     # get pid so we can send SIGINT to it specifically
     pid = os.getpid()
@@ -169,6 +170,7 @@ def test_sigint_RE(RE_abort):
     assert RE_abort._exit_status == 'success'
 
 
+@skip_if_win32_generic
 def test_sigquit_two_hits(RE_abort):
     import time
 
--- a/hutch_python/tests/test_utils.py
+++ b/hutch_python/tests/test_utils.py
@@ -1,7 +1,15 @@
+import asyncio
 import logging
+import os
+import threading
+import time
+from signal import SIGINT
 
+import bluesky.plan_stubs as bps
 import IPython.lib.pretty as pretty
 import pytest
+from bluesky import RunEngine
+from bluesky.utils import RunEngineInterrupted
 
 from hutch_python import utils
 
@@ -119,3 +127,44 @@ def test_hutch_banner():
     logger.debug('test_hutch_banner')
     utils.hutch_banner()
     utils.hutch_banner('mfx')
+
+
+@pytest.fixture(scope='function')
+def RE_abort():
+    loop = asyncio.new_event_loop()
+    loop.set_debug(True)
+    RE = RunEngine({}, loop=loop,
+                   context_managers=[utils.AbortSigintHandler])
+
+    yield RE
+
+    if RE.state != 'idle':
+        RE.halt()
+
+
+def test_abort_RE(RE_abort):
+    def wait_plan():
+        # arbitrarily long wait time
+        yield from bps.sleep(10)
+
+    # get pid so we can send SIGINT to it specifically
+    pid = os.getpid()
+
+    def sigint_signal():
+        time.sleep(2)
+        os.kill(pid, SIGINT)
+
+    # send sigint in a different thread
+    thread = threading.Thread(target=sigint_signal)
+    thread.daemon = True
+    thread.start()
+
+    try:
+        RE_abort(wait_plan())
+    except RunEngineInterrupted:
+        # we expect to interrupt the RunEngine
+        # other exceptions should break the test
+        pass
+
+    assert RE_abort.state == 'idle'
+    assert RE_abort._exit_status == 'abort'
--- a/hutch_python/tests/test_utils.py
+++ b/hutch_python/tests/test_utils.py
@@ -2,13 +2,15 @@ import asyncio
 import logging
 import os
 import threading
-import time
-from signal import SIGINT
+import time as ttime
+from signal import SIGINT, SIGQUIT
 
 import bluesky.plan_stubs as bps
 import IPython.lib.pretty as pretty
 import pytest
 from bluesky import RunEngine
+from bluesky.plan_stubs import abs_set
+from bluesky.preprocessors import finalize_wrapper
 from bluesky.utils import RunEngineInterrupted
 
 from hutch_python import utils
@@ -131,10 +133,12 @@ def test_hutch_banner():
 
 @pytest.fixture(scope='function')
 def RE_abort():
+    """ RunEngine with extra context_managers """
     loop = asyncio.new_event_loop()
     loop.set_debug(True)
     RE = RunEngine({}, loop=loop,
-                   context_managers=[utils.AbortSigintHandler])
+                   context_managers=[utils.AbortSigintHandler,
+                                     utils.SigquitHandler])
 
     yield RE
 
@@ -142,7 +146,7 @@ def RE_abort():
         RE.halt()
 
 
-def test_abort_RE(RE_abort):
+def test_sigint_RE(RE_abort):
     # get pid so we can send SIGINT to it specifically
     pid = os.getpid()
 
@@ -151,7 +155,7 @@ def test_abort_RE(RE_abort):
         yield from bps.sleep(10)
 
     def sim_kill():
-        time.sleep(0.05)
+        ttime.sleep(0.05)
         os.kill(pid, SIGINT)
 
     # send sigint in a different thread
@@ -163,3 +167,41 @@ def test_abort_RE(RE_abort):
 
     assert RE_abort.state == 'idle'
     assert RE_abort._exit_status == 'success'
+
+
+def test_sigquit_two_hits(RE_abort):
+    import time
+
+    from ophyd.sim import motor
+    motor.delay = .5
+
+    pid = os.getpid()
+
+    def sim_kill(n):
+        for j in range(n):
+            time.sleep(.05)
+            os.kill(pid, SIGQUIT)
+
+    lp = RE_abort.loop
+    motor.loop = lp
+
+    def self_sig_int_plan():
+        # three hits will quit pytest
+        threading.Timer(.05, sim_kill, (2,)).start()
+        yield from abs_set(motor, 1, wait=True)
+
+    start_time = ttime.time()
+    with pytest.raises(RunEngineInterrupted):
+        RE_abort(finalize_wrapper(self_sig_int_plan(),
+                                  abs_set(motor, 0, wait=True)))
+    end_time = ttime.time()
+
+    assert RE_abort.state == 'paused'
+    # not enough time for motor to cleanup, but long enough to start
+    assert 0.05 < end_time - start_time < 0.3
+    RE_abort.abort()  # now cleanup
+
+    done_cleanup_time = ttime.time()
+    # this should be 0.5 (the motor.delay) above, leave sloppy for CI
+    assert 0.3 < done_cleanup_time - end_time < 0.6
+    assert RE_abort.state == 'idle'
--- a/hutch_python/utils.py
+++ b/hutch_python/utils.py
@@ -480,31 +480,33 @@ class SigquitHandler(SignalHandler):
         # Check for pause requests from keyboard.
         # TODO, there is a possible race condition between the two
         # pauses here
-        if self.RE.state.is_running and (not self.RE._interrupted):
-            if (self.last_sigint_time is None or
-                    time.time() - self.last_sigint_time > 10):
-                # reset the counter to 1
-                # It's been 10 seconds since the last SIGQUIT. Reset.
-                self.count = 1
-                if self.last_sigint_time is not None:
-                    logger.debug("It has been 10 seconds since the "
-                                 "last SIGQUIT. Resetting SIGQUIT "
-                                 "handler.")
-                # weeee push these to threads to not block the main thread
-                threading.Thread(target=self.RE.request_pause,
-                                 args=(True,)).start()
-                print("A 'deferred pause' has been requested. The "
-                      "RunEngine will pause at the next checkpoint. "
-                      "To pause immediately, hit Ctrl+\\ again in the "
-                      "next 10 seconds.")
+        if not self.RE.state.is_running or self.RE._interrupted:
+            return
 
-                self.last_sigint_time = time.time()
-            elif self.count == 2:
-                print('trying a second time')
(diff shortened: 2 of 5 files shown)


hutch_python/utils.py
```diff
@@ -433,3 +436,22 @@ def get_fully_qualified_domain_name():
...
+
+
+class AbortSigintHandler(SignalHandler):
+    def __init__(self, RE):
+        super().__init__(signal.SIGINT, log=RE.log)
+        self.RE = RE
+
+    def handle_signals(self):
+        # Check for pause requests from keyboard.
+        if self.RE.state.is_running and (not self.RE._interrupted):
```
klauer, line 455: Interrupted status seems like a thing that external code should be query without poking around with RE internals.

I think that might be a quick bluesky PR to add such a property, if you're up for it...

tangkong:
The threaded test fails on PIP but passes on Conda.  I don't understand but I'm trying to

hutch_python/utils.py
ZLLentz: Is the interrupted check necessary here? What does it do?

hutch_python/utils.py
```diff
@@ -433,3 +436,22 @@ def get_fully_qualified_domain_name():
...
+            threading.Thread(target=self.RE.abort(reason='sigint')).start()
```
ZLLentz, line 456: I think it's worth pausing for a moment to make sure that `abort` is the one we want. After all, we could also have picked `stop` or `halt`.

Semantically, an abort is a failed run, but we don't necessarily know if the run was successful or failed just because it was interrupted early.

Practically, an abort raises an exception that creates a long traceback in the terminal, something a stop does not do.

Any thoughts about this?

klauer: We can't guess the user's intentions, of course. In the interest of avoiding tracebacks which will be the subject of many complaints, `stop` seems appropriate.

tangkong: I guess I didn't balk at the ~8 lines of exception traceback, though getting rid of that is certainly nice.  

I generally wouldn't consider any interrupted run "successful", even if intentional.  Exit statuses aren't recorded anywhere by the daq, so the distinction between `stop` and `abort` might just be a superficial one.

`halt` doesn't do any cleanup, so I'd shy away from that option.

ZLLentz: I think the the thing that bugs me about the abort traceback is that it's pretty useless- is it ever different?
Agree on `halt` not being a good choice

tangkong: I can get onboard with `stop`.   The traceback from the `RunEngineInterrupted` exception tends to be descriptive enough.

klauer: As Zach noted it's almost always the same - it's not as if we're catching KeyboardInterrupt at a random point, the RE exits at a consistent spot and yields back control.

So I'd then say it's of benefit to no one to even log the traceback, let alone print it to the user.

hutch_python/utils.py
tangkong: Ah that may be vestigial.  The old RE could be in an interrupted and running state if a plan was paused.  I think the way we have it now, that shouldn't occur (we interrupt and immediately reset the RE).  If we find a way to capture a different signal for pausing the plan, this might be necessary though

tangkong:
> The threaded test fails on PIP but passes on Conda. I don't understand but I'm trying to

We found that the tests fail due to mpl backend issues, this was fixed in [pcds-ci-helpers](https://github.com/pcdshub/pcds-ci-helpers/pull/70), so changes to the travis settings were reverted.

hutch_python/utils.py
```diff
@@ -433,3 +436,22 @@ def get_fully_qualified_domain_name():
...
+
+class AbortSigintHandler(SignalHandler):
+    def __init__(self, RE):
+        super().__init__(signal.SIGINT, log=RE.log)
+        self.RE = RE
+
+    def handle_signals(self):
+        # Check for pause requests from keyboard.
+        if self.RE.state.is_running:
+            threading.Thread(target=self.RE.stop()).start()
```
ZLLentz, line 456: Does this need to be in a thread? I ask because `self.RE.stop()` is called directly here before the thread starts, the object passed into the `target` here is actually the return value from `RE.stop()`

tangkong: Good catch.  I was too lazy in mimicking the existing SIGINT handler, which actually needs the threading.

klauer: This is one more character than yesterday's comma-at-EOL one, but in my opinion so much easier to gloss over. Very nice catch.

hutch_python/tests/test_utils.py
```diff
@@ -119,3 +127,39 @@ def test_hutch_banner():
...
+    RE = RunEngine({}, loop=loop,
+                   context_managers=[utils.AbortSigintHandler])
+
+    yield RE
+
+    if RE.state != 'idle':
+        RE.halt()
+
+
+def test_abort_RE(RE_abort):
```
ZLLentz, line 145: Tiny detail here, `SIGINT` tests using threaded `os.kill` calls don't work on windows (they terminate the entire test suite...)
https://docs.python.org/3/library/os.html#os.kill
I don't actually know any other ways to test this, I've been skipping these to move on with my life:
```suggestion
@skip_if_win32_generic
def test_abort_RE(RE_abort):
```

ZLLentz (approved):
This ended up pretty small, but these are hard-fought lines of code. I think this is implemented in the cleanest way based on the initial request and the available APIs.

hutch_python/tests/test_utils.py
klauer: Naive guess since I haven't looked into details of this since bluesky-2015, but I'd imagine `raise KeyboardInterrupt` in the plan wouldn't be the same, would it?

hutch_python/utils.py
```diff
@@ -433,3 +435,22 @@ def get_fully_qualified_domain_name():
         )
         logger.debug("getfqdn failed", exc_info=True)
         return ""
+
+
+abort_msg = """
+Your RunEngine has been interrupted.  If you aborted the current run (Ctrl+C),
+the RunEngine is ready for use.  If you paused the run, you must resume,
```
klauer, line 442: We don't allow for pausing at the moment, so maybe we shouldn't mention it just yet?

hutch_python/utils.py
```diff
@@ -433,3 +435,22 @@ def get_fully_qualified_domain_name():
...
+class AbortSigintHandler(SignalHandler):
+    def __init__(self, RE):
+        super().__init__(signal.SIGINT, log=RE.log)
+        self.RE = RE
+
+    def handle_signals(self):
+        # Check for pause requests from keyboard.
+        if self.RE.state.is_running:
+            self.RE.stop()
+            print("Aborting current run.")
```
klauer, line 456: I think this would be good to put through the logger, just to give it a timestamp and everything.
`logger.warning` perhaps?

klauer (approved):

hutch_python/tests/test_utils.py
ZLLentz: I should have documented my previous efforts better. Last time I tried `raise KeyboardInterrupt` and I also tried using the alternate windows-only signal and had some sort of problem each time.

tangkong: I was looking around and saw something about mapping a windows Ctrl+C signal to sigint before hitting the handler, which might be a good way forward.  Do we want to try to support windows proper?  Our hutch environments are all on linux right?

hutch_python/utils.py
tangkong: Very true, this was me being hopeful that I'd figure out the Ctrl+P type of functionality early on.  I'll see if I can't make any headway, otherwise I'll change the message

hutch_python/tests/test_utils.py
ZLLentz: We've been more or less ignoring proper support for Windows, but it's nice for the tests to not break entirely when you're developing something else in the library on a Windows machine.

tangkong:
The tests pass despite failing sometimes on my local.  Both of these sub-tests pass without issue if I run them individually (`pytest test_utils.py`), but fail with a full `pytest` invocation

First there are some inconsistent failures in `test_log_setup.py`, where the logger seems to be picking up extra messages not sent in the test suite.  

The test in `test_utils.py` related to this PR then fail on a timing related statement.  The previous failed tests might be causing some kind of slowdown within pytest?  There's already some slop built into the test to account for CI inconsistency.

hutch_python/utils.py
klauer: Looks like this became relevant again?

tangkong: > Looks like this became relevant again?

Justifying extraneous documentation the hard way

hutch_python/utils.py
```diff
@@ -433,3 +436,75 @@ def get_fully_qualified_domain_name():
...
+        self.RE = RE
+        self.last_sigint_time = None  # time most recent SIGQUIT was processed
+        self.num_sigints_processed = 0  # count SIGQUITs processed
+
+    def __enter__(self):
+        return super().__enter__()
+
+    def handle_signals(self):
+        # Check for pause requests from keyboard.
+        # TODO, there is a possible race condition between the two
```
klauer, line 481: So the race condition is why we check the time to ensure we don't see two `c-\` for one keypress?

klauer (approved):
A few minor things to fix up, I think, but overall I think this is great.

I think I'm a convert as to this being actually a good idea. I like that `ctrl-c` retains its functionality as to what you'd expect in Python. Pausing is, after all, something bluesky-specific and special. While it may be second-nature to us now, it's not intuitive and certainly confusing unless you take the time to read the page-long explanation about `RE.resume()` and whatnot.

That said, I did have to test this outside of tmux, as `ctrl-\` somehow interprets it as a different mapped key combination. I have a feeling that's the fault of my configuration. And of course, we don't have to worry about scientists using hutch-python on tmux.

As to the testing - I think if it works locally, it's probably fine for now in my opinion. Testing this stuff via the test suite is finicky and painful and tough to get right.

hutch_python/utils.py
```diff
@@ -480,31 +480,33 @@ def handle_signals(self):
...
-                logger.debug("RunEngine detected two SIGQUITs. "
-                             "A hard pause will be requested.")
-
-                threading.Thread(target=self.RE.request_pause,
-                                 args=(False,)).start()
+        if not self.RE.state.is_running or self.RE._interrupted:
+            return
+
+        if (self.last_sigint_time is None or
+                time.time() - self.last_sigint_time > 10):
```
klauer, line 487: Sorry, one last thing: generally, you should use `time.monotonic()` which will increase correctly - regardless of weird things like leap seconds or system/NTP clock changes - in place of `time.time()` (which has no such guarantee)

hutch_python/utils.py
```diff
@@ -433,3 +436,75 @@ def get_fully_qualified_domain_name():
...
+                threading.Thread(target=self.RE.request_pause,
+                                 args=(True,)).start()
+                print("A 'deferred pause' has been requested. The "
+                      "RunEngine will pause at the next checkpoint. "
+                      "To pause immediately, hit Ctrl+\\ again in the "
+                      "next 10 seconds.")
+
+                self.last_sigint_time = time.time()
+            elif self.count == 2:
+                print('trying a second time')
```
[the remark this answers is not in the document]
klauer: Weird... I think we could at least do better message-wise...

hutch_python/utils.py
```diff
@@ -433,3 +436,75 @@ def get_fully_qualified_domain_name():
...
+                # It's been 10 seconds since the last SIGQUIT. Reset.
+                self.count = 1
+                if self.last_sigint_time is not None:
+                    logger.debug("It has been 10 seconds since the "
+                                 "last SIGQUIT. Resetting SIGQUIT "
+                                 "handler.")
+                # weeee push these to threads to not block the main thread
+                threading.Thread(target=self.RE.request_pause,
+                                 args=(True,)).start()
+                print("A 'deferred pause' has been requested. The "
```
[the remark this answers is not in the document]
tangkong: I'm a fan of reminding the user what the behavior is at the first attempt to pause, but you're right that it'd be nice to have this in the logs.  I'll add a logger line in addition

hutch_python/utils.py
tangkong: That's my understanding, yes.

hutch_python/utils.py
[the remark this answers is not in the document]
klauer: Hmm, `logger.warning` should get to the user as well, shouldn't it?

hutch_python/utils.py
```diff
@@ -433,3 +436,77 @@ def get_fully_qualified_domain_name():
...
+        # Check for pause requests from keyboard.
+        if self.RE.state.is_running:
+            self.RE.stop()
+            # Use our own logging
+            logger.warning("Aborting current run.")
+
+
+class SigquitHandler(SignalHandler):
+    """
+    Interprets SIGQUIT (Ctrl+\\) as standard RunEngine interrupt behavior.
```
ZLLentz, line 466: This seems reasonable from a user/usage perspective, but confusing from a linux perspective. TIL that `SIGQUIT` is the same as `SIGINT` except that the process is supposed to produce a core dump. On balance I'd say this is a good approach but I'm a bit wary.

hutch_python/utils.py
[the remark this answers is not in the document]
tangkong: You're totally right, I spaced there.  `logger.warning` is great.

hutch_python/utils.py
klauer: `SIGQUIT` seems to be caught by IPython and exits _without_ a core file, as far as I can tell.

If the concern is strong enough, we could look into what it'd take to hook all keyboard events to filter for the original proposed `Ctrl-P`. 

I guess it'd be good to fully figure out the level of concern prior to going forward, as we wouldn't want to switch out the key combination from under the users down the line.

ZLLentz: My concern isn't that high of any practical problems arising from this. This seems like the most efficient way to get to the goal of "some keypress to retain the pausing behavior while letting ctrl+c feel like stopping the run entirely". It just seems semantically discordant.

As a side note, understanding good approaches for responding to "any" keypress during something like a scan might be a good thing for us to learn.

tangkong: I guess one concern is that `SIGQUIT` will exit ipython if the RunEngine is not going through the pause routine.  I can see that causing grief in the future if users hit `Ctrl-\` three times and have to re-load hutch-python.  `SIGINT` was really the signal for the job.  

You could argue that our users should be smashing `Ctrl-\` if `Ctrl-c` already aborts, but that might be a risky assumption

klauer: I guess I was too focused on how nicely this worked in testing (and didn't mash the keys as you might expect users to). I think @tangkong 's point is strong enough to warrant finding an alternative signal - or better, a keyboard hook.

This was also enough to make me dig into the details of keybindings, and it may be easier than I had assumed: https://github.com/ipython/ipython/blob/3587f5bb6c8570e7bbb06cf5f7e3bc9b9467355a/IPython/terminal/shortcuts.py#L58

And even possible without an additional library...

tangkong: Turns out this doesn't work, as the RunEngine thread prevents input from reaching the ipython prompt_toolkit.  

A polling thread watching for a separate key combination has been suggested by Zach multiple times.  I'll look into it a bit.

tangkong:
In further testing, I wasn't able to hit ctrl-\ fast enough to cause ipython to quit, once the modules were loaded the RunEngine seemed to be able to abort fast enough.  This isn't a firm guarantee, and I imagine it'd depend on the plan and hardware possibly.  

If we do end up adding a polling layer, I might argue it would be worth its own PR.  It'd end up being pretty sizable I think?  (maybe I'm also done looking at this PR haha)

klauer:
It would have been nice to get that on-startup IPython code linted or in some helper function, but I don't see an easy hook for it. Best alternative could be to execute `import hutch_python; hutch_python._ipython_startup_hook_()` perhaps instead of the block of code?

... but I'd rather not hold this PR up any further. My opinion is that this is good enough for a feature that will not likely be used.

tangkong:
Yea I struggled with where to put that little block of code, the issue being I can only unbind `c-\\` once ipython has initialized, and we don't actually run startup files anywhere.  

Is `hutch_python._ipython_startup_hook_()` something that already exists? I can't seem to find it

klauer:
> Is hutch_python._ipython_startup_hook_() something that already exists? I can't seem to find it

No, it'd be something new, including all the code from [825a159](https://github.com/pcdshub/hutch-python/pull/324/commits/825a15938831540f8f37a4c3c98a0c8ac476d97a)

klauer:
One more time

ZLLentz:
Ok I think this is how we want it now- we can merge, right? Some follow-up task at "generic future date" for a startup hook?

tangkong:
Thanks for the feedback everyone!  Appreciate all the help along the way! 😄

PR #327, opened 2022-03-29
ENH: output environment info at startup

tangkong:
## Description
Print environment information at hutch python startup.  

## Motivation and Context
With the variation in environments used around the facility, it'd be good to expose exactly what's being used when hutch-python is started.  

## How Has This Been Tested?
Interactively.

In this PR

## Screenshots (if appropriate):

image

## Pre-merge checklist

### DIFF
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -90,9 +90,13 @@ def get_env_info():
 
     banner = (
         'Environment Information\n'
-        f'  conda env: {conda_ver}\n'
-        f'  dev pkgs: {dev_pkgs}'
+        f'  Conda Environment: {conda_ver}\n'
+        f'  Development Packages: {" ".join(dev_pkgs)}'
     )
+
+    logger.info(
+        f'Conda Environment: {conda_ver}, '
+        f'Development Packages: {" ".join(dev_pkgs)}')
     return banner
 
 
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -69,6 +69,33 @@ def configure_tab_completion(ipy_config):
         IPython.core.completer.dir2 = dir
 
 
+def get_env_info():
+    """
+    Gather python environment info.
+
+    Returns
+    -------
+    banner : str
+        String with conda environment, dev packages
+    """
+    banner = ""
+
+    # grab activated conda env
+    conda_ver = os.environ.get('CONDA_DEFAULT_ENV', 'None')
+    dev_pkgs = [
+        os.path.basename(os.path.dirname(p))
+        for p in os.environ.get('PYTHONPATH', 'None').split(":")
+        if p
+    ]
+
+    banner = (
+        'Environment Information\n'
+        f'  conda env: {conda_ver}\n'
+        f'  dev pkgs: {dev_pkgs}'
+    )
+    return banner
+
+
 def configure_ipython_session():
     """
     Configure a new IPython session.
@@ -105,6 +132,10 @@ def configure_ipython_session():
         "cb = Keys.ControlBackslash; "
         "ip.pt_app.key_bindings.remove(cb) "
     )
+
+    # modify ipython banner
+    ipy_config.TerminalInteractiveShell.banner2 = get_env_info()
+
     return ipy_config
 
 
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -16,6 +16,7 @@ from IPython import start_ipython
 from traitlets.config import Config
 
 from .constants import CONDA_BASE, DIR_MODULE
+from .env_version import get_env_info, log_env
 from .load_conf import load
 from .log_setup import configure_log_directory, debug_mode, setup_logging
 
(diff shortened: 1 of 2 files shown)


hutch_python/cli.py
```diff
@@ -69,6 +69,33 @@ def configure_tab_completion(ipy_config):
...
+    conda_ver = os.environ.get('CONDA_DEFAULT_ENV', 'None')
+    dev_pkgs = [
+        os.path.basename(os.path.dirname(p))
+        for p in os.environ.get('PYTHONPATH', 'None').split(":")
+        if p
+    ]
+
+    banner = (
+        'Environment Information\n'
+        f'  conda env: {conda_ver}\n'
```
klauer, line 93:
```suggestion
        f'  Conda Environment: {conda_ver}\n'
        f' Development packages: {" ".join(dev_pkgs)}'
```

klauer (commented):
Looks nice to me.

I'd imagine this could also go somewhere in the `load_conf` and target a log message, but I'm not sure it would be any better than what you are proposing. Perhaps only that the order of operations would be clearer, or less obfuscated behind `IPython` settings? 🤷

ZLLentz:
ah shoot we doubled up on this

hutch_python/env_version.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def get_standard_dev_pkgs() -> set[str]:
+    """Check the standard dev package locations for hutch-python"""
+    pythonpath = os.environ.get('PYTHONPATH', '')
+    if not pythonpath:
+        return set()
+    pkg_names = set()
+    for part in pythonpath.split(':'):
+        if any([(s in part) for s in _dev_ignore_list]) or (not part):
```
klauer, line 56: Unnecessary list comprehension:
```suggestion
        if any(((s in part) for s in _dev_ignore_list)) or (not part):
```

hutch_python/env_version.py
```diff
@@ -0,0 +1,73 @@
...
+            continue
+        pkg_names.add(os.path.basename(os.path.dirname(part)))
+    return pkg_names
+
+
+def get_env_info() -> str:
```
klauer, line 62: Missing docstring

klauer (commented):
Looks good overall 👍

hutch_python/env_version.py
tangkong: Is the concern that we don't want to use a mutable list where we don't have to?  Apparently list comprehensions are faster than tuple comprehensions (at least in my timing tests).

Within this function there's a minimal speed difference between tuple and list comprehensions.

klauer: It's entirely negligible as far as speed, and mutability isn't relevant. It's rather just a generally good pattern to follow.

What is the actual difference then? A temporary list is generated in the original code. In my suggestion, it's only a generator which works on an item-by-item basis - no tuple is generated, even though it may look like that syntactically. Make more sense? (Replying on mobile at the moment, otherwise I'd throw in some code as well)

I should find a better way of marking such suggestions - but laziness/brevity won out for my comment.

tangkong: Thanks for the clarification, makes perfect sense.

hutch_python/env_version.py
```diff
@@ -0,0 +1,73 @@
...
+
+import logging
+import os
+import os.path
+
+import pkg_resources
+
+logger = logging.getLogger(__name__)
+
+_dev_ignore_list = ['x86_64-rhel7-opt', 'x86_64-linux-opt']
```
ZLLentz, line 14: If the goal is to cut out the daq lib spam, there a few more architectures to pick here
```
$ ls /reg/g/pcds/dist/pds/current/build/pdsapp/lib/
i386-linux-dbg  i386-linux-opt  x86_64-linux  x86_64-linux-dbg  x86_64-linux-opt  x86_64-rhel7  x86_64-rhel7-dbg  x86_64-rhel7-opt
```
or we could also switch the targets to `pdsapp` and `ami`, maybe? To specifically exclude the control and monitoring softwares respectively?

ZLLentz (approved):
This is a good compromise/merging of the opposing needs of "log the env information early for debug" and "show the env information in the startup banner so the user actually sees it".

I couldn't figure out how to make unit tests for these because I couldn't figure out how to change environments during a pytest run...

hutch_python/env_version.py
tangkong: I didn't know the proper targets here, good to know.

hutch_python/env_version.py
```diff
@@ -0,0 +1,73 @@
...
+        logger.debug(
+            'Using conda env %s with no dev packages',
+            get_conda_env_name(),
+        )
+    logger.debug(dump_env())
+
+
+def dump_env() -> list[str]:
+    """
+    Get all packages nad versions from the current environment.
```
klauer, line 36:
```suggestion
    Get all packages and versions from the current environment.
```

klauer (approved):
If you wanted to do unit tests, the easiest way would be to just mock a few of these supporting functions and test the rest, I think. (pytest fixture `monkeypatch` and `monkeypatch.setattr(obj, attr, test_replacement)`)

ZLLentz:
What you really want to test is that the various functions give the correct values though- and the best way to get those correct values is to rewrite the function again... but in the test suite. How would you catch a regression like conda changing which env variable it uses to store your conda env?

klauer:
> What you really want to test is that the various functions give the correct values though- and the best way to get those correct values is to rewrite the function again... but in the test suite. How would you catch a regression like conda changing which env variable it uses to store your conda env?

You're 100% right, @ZLLentz. 

In the land of testing, sometimes there are compromises in order to be time efficient. With (1) being ideal to (3) being least ideal:
1. Spend a ton of time to ensure all of your functions are 100% correct and covered
2. Spend a small amount of time to ensure - that given some reasonable assumptions about the underlying environment - that user-facing functions will continue to work
4. Spend no time and do no testing outside of running it to see if it does what you expect

My suggestion was to sit around the (2) mark, but I think in this case (3) is reasonably acceptable, too.

ZLLentz:
I think it's time to bring this in!

PR #328, opened 2022-03-29
ENH: add conda env name, dev pkgs, and debug env dump at startup

ZLLentz:
## Description

Same as title
```
INFO     log_env - Using conda env dev with dev packages hutch_python, pcdsdevices
```
This will have `conda env pcds-5.4.0` or whatever is being used instead of `conda env dev` for someone using the normal environments.

## Motivation and Context

To help users understand what exactly is being run.

## How Has This Been Tested?

Interactively only so far

Not yet


ZLLentz:
closing in favor of #327

PR #330, opened 2022-04-06
ENH: add post-IPython-initialization routines, better banner

### LINKED ISSUE #326: Startup hook for running code after IPython initialization
## Expected Behavior
A thought that came up during #324 , we should provide a hook to run code after ipython initialization.  IPython normally accomplishes this with startup files, but in our case we'd probably want to dictate this through `conf.yml` files or even external files.  

## Current Behavior
Currently we initialize the ipython / hutch-python session via traitlets configuration files.  

In the aforementioned PR, we utilized the ipython configuration attribute `InteractiveShellApp.exec_lines` to remove a keybinding, but this code isn't linted in any way (and is kind of ugly to look at)

This is distinct from the way we currently initialize ipython, which is mostly setting up object and passing them to the ipython namespace.  

## Possible Solution
Perhaps a conf.yml field that points at files to execute or a submodule to run.  

## Steps to Reproduce (for bugs)
N/A

## Context
Most recently, trying to unbind an ipython keybind.  In the future, maybe this becomes more generally useful.  

## Your Environment
pcds-5.3.0

tangkong:
## Description
Configures the following settings:
- ELogPoster Callback subscription to RunEngine
- Unbinding `Ctrl+\\` sigquit handler

Disables default banner and makes a new banner with helpful namespace hints.  In so doing:
- Modifies LoadCache to write also doc information to a `._desc` attribute to store a short description

Also fixes a bug in the default groups section of `load_conf`, where if a default class namespace was not populated (eg. `motors` in a sim session), default groups would fail to load because a docstring could not be assigned.  `LoadCache` now only assigns a docstring if the object exists.  This fixes the bug where "all objects" would lack its docstring, since the this bug would throw an exception before its docstring was assigned.

## Motivation and Context
Closes #326 

A few desired configuration settings require modifying configuration of IPython after it has been initialized.  Previously there was no place for this.  

## How Has This Been Tested?
tests pass locally.  Interactive testing

image

This PR

## Pre-merge checklist
- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page


hutch_python/_startup_script.py
```diff
@@ -0,0 +1,40 @@
...
+    logger.debug(exc, exc_info=True)
+
+try:
+    # RE, ELog will already exist by now
+    elogc = ELogPoster(elog, IPython.get_ipython())  # noqa
+    RE.subscribe(elogc)  # noqa
+except Exception as exc:
+    logger.debug(exc, exc_info=True)
+
+# Disable auto-suggestions.  Currently no config option exists.
```
ZLLentz, line 36: Was there ever a consensus on this?

hutch_python/_startup_script.py
```diff
@@ -0,0 +1,40 @@
+"""
```
ZLLentz, line 1: I'm glad this is pulled out into a lintable file, but after ruminating a bit I don't think that any of these need to be deferred until the exec_lines hook. Can't all of these be ipython extensions instead?

What are the advantages of doing anything library-side in the exec_lines?

I think I have some built-in skepticism whenever I see strings being exec'd

hutch_python/cli.py
```diff
@@ -98,14 +134,9 @@ def configure_ipython_session():
...
-    ipy_config.InteractiveShellApp.exec_lines = (
-        "import IPython; "
-        "from prompt_toolkit.keys import Keys; "
-        "ip = IPython.get_ipython(); "
-        "cb = Keys.ControlBackslash; "
-        "ip.pt_app.key_bindings.remove(cb) "
-    )
+    # Run startup hook code
+    lines = get_startup_hook(cfg=cfg)
+    ipy_config.InteractiveShellApp.exec_lines = lines
```
ZLLentz, line 139: I've managed to confuse myself reviewing this PR- what specific features are we trying to enable for the user via adding a post-init hook?

ZLLentz (commented):
I'm not sure if I've just forgotten something important, but I've lost track of the thread/idea behind this. The issue doesn't jog my memory either.

hutch_python/_startup_script.py
tangkong: I remember there being a [ticket](https://jira.slac.stanford.edu/browse/LCLSECSD-723) where a few people expressed distaste for the feature, but we never properly polled the populous, no.

hutch_python/_startup_script.py
tangkong: Yea I'm not a fan of exec-ing stuff either, and I seem to be doing it a lot..

The ipython extensions we set up in the traitlets config cannot access the namespace items (devices, RE, etc).  That's really the primary reason for this functionality.  Things like the elog callback require the elog instance as well as the ipython input record for initialization.  

I'll admit I overlooked the ipython extensions earlier, and 2/3 of the configurations I singled out can be moved there

hutch_python/cli.py
tangkong: My target has shifted to subscribing the elog callback specifically.  I thought giving users the ability to run something with their devices on hutch-python startup might be useful, but I'm also ok hiding that functionality if we think it crosses a line hutch-python shouldn't.

hutch_python/_startup_script.py
ZLLentz: Gotcha, there's no way at all to pass resources into IPython extensions, except gross globals, maybe? That's a bummer.

hutch_python/cli.py
ZLLentz: I don't think it crosses a line per se I just failed to be imaginative enough to figure out what it would be used for. I think a lot of things can be done in `beamline.py`.

klauer: Couldn't this be a minimal import-and-call like the following?

`import hutch_python; hutch_python.run_startup_hook()`

This gives the benefit of:
1. Lintable and checkable file as in this PR
2. Only 2 lines of source code that come without traceability and are extremely unlikely to fail (and if they do, we have bigger problems)
3. Much less stringified code compared to this PR

hutch_python/_startup_script.py
```diff
@@ -0,0 +1,40 @@
...
+import logging
+
+import IPython
+from nabs.callbacks import ELogPoster
+from prompt_toolkit.keys import Keys
+
+logger = logging.getLogger(__name__)
+
+# Unbind Ctrl+\\
+try:
```
klauer, line 21: Should these use the `safe_load` context manager to give a description of what it's doing?

klauer, line 21: Should these use the `safe_load` context manager to give a description of what it's doing?

klauer (commented):
I think this could be a useful addition but think we should spend a bit more time solidifying it.

For each of these settings where we don't have a consensus yet, maybe a `conf.yml` setting would be appropriate. It may be a bit weird that these are duplicated from the users' IPython configuration files, though.

As to "why not just beamline.py?" I think that whatever configuration we perform in this file is something that will be globally applied to all hutches (as in our normal `load_conf()`) but requires access to the running IPython instance/user namespace.

Were we to have the hook I mentioned above (`hutch_python.run_startup_hook()`) an exec-less middle-ground solution would be to add that call in every `beamline.py`. (I think I prefer the automatic version that hutch-python performs, personally, as it's less error-prone)

hutch_python/_startup_script.py
tangkong: duplicate, resolving

tangkong: This was mostly cosmetic, as `exec_lines` and ipython extensions happen after the banner is printed 

image

Maybe it's worth re-doing the banner in one of these extensions to clean this up.  I thought the user would be confused if there suddenly safe_load messages after the banner

tangkong:
Made some pretty significant changes, thanks for the feedback everyone.  
I've narrowed the scope of this to focus on things I think we have more consensus on.  This has turned into less of a startup-hook, but that's ok with me.

hutch_python/cli.py
```diff
@@ -98,17 +99,15 @@ def configure_ipython_session():
...
-    ipy_config.TerminalInteractiveShell.banner2 = get_env_info()
+    # disable default banner
+    ipy_config.TerminalIPythonApp.display_banner = False
+
+    # Run startup hook code, print banner after startup hook files
+    files = [
+             str(Path(__file__).parent / 'startup_script.py'),
+             str(Path(__file__).parent / 'print_hint_banner.py')
+            ]
+    ipy_config.InteractiveShellApp.exec_files = files
```
klauer, line 110: What happens if you do `hutch_python my_script.py`; does that go to `exec_files` or elsewhere?

tangkong: Good point.  The IPython shell wouldn't be initialized, so the entire `configure_ipython_session` block (and `exec_files`) would be ignored entirely.  

I suppose if someone's script requires the ELogPoster callback, the script wouldn't work

klauer: It probably doesn't matter in the end, but I was curious if that was the same mechanism.

ZLLentz: fwiw I suspect this script option is under-utilized due to slow startup speed

hutch_python/cli.py
```diff
@@ -98,17 +99,15 @@ def configure_ipython_session():
...
-    )
-
-    # add env info to ipython banner
-    ipy_config.TerminalInteractiveShell.banner2 = get_env_info()
+    # disable default banner
+    ipy_config.TerminalIPythonApp.display_banner = False
+
+    # Run startup hook code, print banner after startup hook files
+    files = [
+             str(Path(__file__).parent / 'startup_script.py'),
```
klauer, line 107: We apparently have a constant for this path already -
```suggestion
             str(DIR_MODULE / 'startup_script.py'),
```

hutch_python/print_hint_banner.py
```diff
@@ -0,0 +1,52 @@
...
+
+    out = ''
+    for k in ns:
+        out += f"  {k} - {getattr(global_ns[k], '_desc', 'N/A')}\n"
+
+    return out
+
+
+base_banner = (
+    "-----------------------------------\n"
```
klauer, line 41: No need to refactor, but multi-line f-strings are a thing:
```python
f"""
------------
{get_env_info()}
...
"""
```

hutch_python/print_hint_banner.py
```diff
@@ -0,0 +1,52 @@
...
+default_objects = ['RE', 'daq', 'elog', 'archive']
+
+
+def gather_hint_table(namespace: List[str]) -> str:
+    """
+    Gather variable name and short description into a table if the
+    variable name is in the current global namespace
+    """
+    global_ns = globals()
+    ns = [x for x in namespace if x in global_ns.keys()]
```
klauer, line 31:
```suggestion
    ns = [x for x in namespace if x in global_ns]
```

hutch_python/startup_script.py
```diff
@@ -0,0 +1,23 @@
...
+imported from.
+"""
+
+import IPython
+from nabs.callbacks import ELogPoster
+
+from hutch_python.utils import safe_load
+
+if __name__ == '__main__':
+    with safe_load('ELogPoster'):
```
klauer, line 20: (It's OK as-is, just a suggestion)

Seeing non-trivial blocks inside of `if __name__ == "__main__"` always makes me a bit uneasy, though.
To that end, I think this could still go in a function, then it could be made more explicit what's to be part of the global namespace:

```python
def _configure_elog_poster():
    with safe_load(...):
        elog = globals().get("elog", None)
        elogc = ...
    return elogc

if __name__ == "__main__":
   elogc = _configure_elog_poster()
```

klauer (approved):
I think the output is pretty great!

My review has a few ideas/thoughts that you can safely ignore.

hutch_python/startup_script.py
klauer: I also meant to include the imports inside the function here so they don't get brought into the session as well (unless they're already there somehow).

hutch_python/print_hint_banner.py
```diff
@@ -0,0 +1,52 @@
...
+    "-----------------------------------\n"
+    f'Helpful Namespaces:\n'
+    f'{gather_hint_table(default_namespaces)}\n'
+    f'Useful objects:\n'
+    f'{gather_hint_table(default_objects)}\n'
+)
+
+
+if __name__ == '__main__':
+    sys.stdout.write(base_banner)
```
ZLLentz, line 52: does `print` behave weirdly in the `exec_files` context?

ZLLentz:
This looks really snazzy now, and you've also fixed some subtle bugs. I'm going to take it for a spin.

hutch_python/startup_script.py
```diff
@@ -0,0 +1,23 @@
...
+
+Code here will take executed after both IPython has been loaded and
+the namespace has been populated with hutch objects.
+
+These will be run as standalone python files, and should not be
+imported from.
+"""
+
+import IPython
+from nabs.callbacks import ELogPoster
```
ZLLentz, line 15: We need to update our `nabs` dependency, the `ELogPoster` is in `v1.3.0`
https://github.com/pcdshub/hutch-python/blob/5f77e01ee28f9f59446aea738215c862c63a6e3d/conda-recipe/meta.yaml#L31

ZLLentz:
Funny output, at no fault of this PR:
```
-----------------------------------
Environment Information
  Conda Environment: dev
  Development Packages: zlentz
-----------------------------------
```

hutch_python/pt_app_config.py
```diff
@@ -0,0 +1,21 @@
...
+from prompt_toolkit.keys import Keys
+
+from .utils import safe_load
+
+logger = logging.getLogger(__name__)
+
+
+def load_ipython_extension(ipython):
+    # Unbind Ctrl+\\
+    with safe_load('disable ctrl+\\, auto suggestions'):
```
klauer, line 18: Auto suggestions are still showing up for me (which is fine for me personally as I think they're useful...):

image

ZLLentz:
Yeah I think there's some adjustment needed to the dev package finder- can be separate from this PR and I'm happy to help look at it. Here's using XPP's pythonpath
```
In [12]: os.environ['PYTHONPATH'] = '/cds/group/pcds/pyps/apps/hutch-python/xpp:/cds/group/pcds/pyps/apps/hutch-python/xpp/dev/devpath:/cds/group/pcds/pyps/apps/hutch-python/xpp/../common/dev
    ...: /devpath:/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/ami-current/build/ami/lib/x86_64-rhel7-opt'

In [13]: get_standard_dev_pkgs()
Out[13]: {'dev', 'hutch-python'}
```
I believe the correct answer here is `pcdsdevices`, `hutch-python`, `archapp` but I'm not 100% on that

hutch_python/pt_app_config.py
tangkong: I took that part out of this PR, since there was some uncertainty as to whether or not we wanted it.  

Previously, I think a few scientists wanted it gone because there was uncertainty around how to ignore the suggestions.  I imagine in time they'll get used to it, but I decided to punt that decision

hutch_python/print_hint_banner.py
tangkong: I so consistently forget you can do this.

hutch_python/print_hint_banner.py
tangkong: No, I was just mimicking how IPython printed the banner.  They seem to have the same end result so for simplicity's sake I'll switch to print.

klauer: They were likely avoiding an additional newline at the end of the block or something along those lines. It's implicit in `print` (controlled by the `end=` kwarg) but not in `sys.stdout.write`

hutch_python/startup_script.py
tangkong: I like this, lets me actually lint those lines.

tangkong:
Updated with the suggestions.  I'll also make an issue for the dev package finder.

hutch_python/pt_app_config.py
klauer:
```suggestion
    with safe_load('disable ctrl+\\'):
```

hutch_python/startup_script.py
ZLLentz: @tangkong did this make it in?

tangkong: Ah I missed it, I'll do that real quick

ZLLentz (approved):
I think an extra newline must have snuck into the environment information banner:
```
-----------------------------------
Environment Information
  Conda Environment: dev
  Development Packages: zlentz

-----------------------------------
Helpful Namespaces:
  a - Namespace of all loaded objects.
  sim - Namespace of simulated hardware.
  bp - Namespace of full bluesky plans.
  re - Wrapped plans for quick running.

Useful objects:
  RE - Bluesky plan executor.
  archive - Archiver interface for getting historical data.
```
I'm laying down my green checkmark regardless, this looks very good

tangkong:
whoops, unintended consequence of me changing how the string was formatted.  Fixed

ZLLentz:
All questions/suggestions addressed, two green checks, output looks good, I am merging the PR

hutch_python/startup_script.py
```diff
@@ -0,0 +1,32 @@
...
+
+    from hutch_python.utils import safe_load
+
+    with safe_load('ELogPoster'):
+        # RE, ELog will already exist by now
+        elog = globals().get('elog', None)
+        RE = globals().get('RE', None)
+        assert elog is not None
+
+        elogc = ELogPoster(elog, IPython.get_ipython())
```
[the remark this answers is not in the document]
tangkong: It won't be, and I don't think there's anything to configure in the callback.  Posts get enabled from the RunEngine call (`RE(plan(), post=True)`).  

Maybe if we add more features to the callback it'll be worth exposing?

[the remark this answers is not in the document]
klauer: Gotcha. I'm not very familiar with that `ELogPoster` instance - a bit of digging on my end would have shown as much (sorry!). This makes more sense now.

No follow-up PRs necessary then. 😥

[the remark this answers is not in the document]
tangkong: To be fair, nobody is yet 😆

Issue #332, opened 2022-04-20
Improve dev package finder


tangkong:
## Expected Behavior
get_standard_dev_pkgs() should return more reasonable list of packages

## Current Behavior
```python 
In [12]: os.environ['PYTHONPATH'] = '/cds/group/pcds/pyps/apps/hutch-python/xpp:/cds/group/pcds/pyps/apps/hutch-python/xpp/dev/devpath:/cds/group/pcds/pyps/apps/hutch-python/xpp/../common/dev
    ...: /devpath:/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/ami-current/build/ami/lib/x86_64-rhel7-opt'

In [13]: get_standard_dev_pkgs()
Out[13]: {'dev', 'hutch-python'}
```

I believe the correct answer here is pcdsdevices, hutch-python, archapp but I'm not 100% on that

## Possible Solution
Probably need to expand the paths similar to how python does.  Naively, one could stop recursively adding directories once an `__init__.py` is found.  But maybe there's a better way to do this.  

## Context
#330 

## Your Environment
pcds-5.3.1

ZLLentz:
I think looking for `__init__.py` was my first thought as well. It's a bit annoying but it'll make the output more thorough.

I think we can just:
- check every directory in the pythonpath (at depth 1)
- if there's an `__init__.py`, that's a module to include
- if there's not such a file, skip it

I guess the main things this misses are compiled c python modules and random .py files? I suppose we can include those too

tangkong:
I would have voted for `setup.py` at first but some of the "packages" were flattened past that point.

I think your suggested procedure should work though

klauer:
I believe the correct/most accurate method here might be to use `pkgutil` - but it's relatively slow. 
"Correct" isn't necessarily right for us, though. I think the `__init__.py` method will likely suffice for what we're after.

<details>

<summary>For reference</summary>

`iter_modules` shows my dev packages pretty readily:

```python
In [26]: pkgutil.iter_modules?
Signature: pkgutil.iter_modules(path=None, prefix='')
Docstring:
Yields ModuleInfo for all submodules on path,
or, if path is None, all top-level modules on sys.path.

'path' should be either None or a list of paths to look for
modules in.

'prefix' is a string to output on the front of every module name
on output.
     ~/miniforge3/lib/python3.8/pkgutil.py
[...]

PR #333, opened 2022-04-28
ENH / STY: improve dev package finder function, run pre-commit-hooks

tangkong:
## Description
A more robust dev package finder, and the style edits required to pass travis

## Motivation and Context
#332 

## How Has This Been Tested?
Tested on my own dev environment, as well as xpp's env.  
pkgutil seems to take only slightly longer than searching for `__init__.py` files

This PR

image


hutch_python/env_version.py
```diff
@@ -51,11 +57,12 @@ def get_standard_dev_pkgs() -> set[str]:
     pythonpath = os.environ.get('PYTHONPATH', '')
     if not pythonpath:
         return set()
-    pkg_names = set()
-    for part in pythonpath.split(':'):
-        if any(((s in part) for s in _dev_ignore_list)) or (not part):
-            continue
-        pkg_names.add(os.path.basename(os.path.dirname(part)))
+    paths = pythonpath.split(':')
```
klauer, line 60: I think `:` should be `os.pathsep`

hutch_python/env_version.py
```diff
@@ -14,6 +15,11 @@
 _dev_ignore_list = ['ami', 'pdsapp']
 
 
+def not_ignored(path, ignores=_dev_ignore_list):
+    """ Return True if _dev_ignore_list isn't in path or empty"""
+    return not (any((s in path for s in ignores)) or not path)
```
klauer, line 20: Did I get this right?
`not (x or not y)` -> `not x and y` -> `y and not x`

```suggestion
    return bool(path and not any(s in path for s in ignores))
```

klauer (approved):
* pre-commit fixes look fine to me
* Couple code smell/things that could be improved (constructive criticism!)

hutch_python/env_version.py
tangkong: logic is hard

ZLLentz:
I like how `xpp` is naturally found as a dev package by this

hutch_python/env_version.py
```diff
@@ -51,11 +57,12 @@ def get_standard_dev_pkgs() -> set[str]:
...
         return set()
-    pkg_names = set()
-    for part in pythonpath.split(':'):
-        if any(((s in part) for s in _dev_ignore_list)) or (not part):
-            continue
-        pkg_names.add(os.path.basename(os.path.dirname(part)))
+    paths = pythonpath.split(':')
+    valid_paths = filter(not_ignored, paths)
+    pkg_names = set(n.name for n in
+                    pkgutil.iter_modules(path=valid_paths)
```
ZLLentz, line 65: I think this was the best possible approach, leveraging the normal python package tools

ZLLentz (approved):
Runs fast and looks very correct

Issue #293, opened 2021-10-07
Per-Device {instance,class} user-configurable settings


klauer:
## Context

There are a certain number of per-`ophyd.Device` configuration settings that users will never be happy with our choice of defaults.  While discussing this, our primary example has been device tab whitelists.

As different users have different (legitimate) needs and preferences, @ZLLentz and I came to the conclusion that it's sensible for us to:
1. Allow the user to configure these without our intervention or a bug report/pull request/release cycle
2. Provide a general framework for these configuration settings.
    - Our goal with the framework should be the ability to support any of the following that we agree would be good to include - even if those aren't added on day 1.

## User-configurable settings

Some potential things that we would add:
- Tab completion settings
- Kind settings (a configuration file overlay, of sorts)
- User interface icons
- Preferred user interface layout or display filename potentially (for `device.screen()`)
- Tweakable device representations (@ZLLentz has an ophyd branch sitting somewhere with Jinja2-based reprs of devices. These templates could be stored in code as defaults and updated as part of config files)
- Device name aliases (then again, rereading https://github.com/pcdshub/hutch-python/issues/213 I somewhat prefer that)
- Preset positions (we already have a supported YAML format for this; I think we will decide to leave it alone)

## Defaults

* We should provide a configuration with sensible defaults. 
    * Good settings shared among many hutches could make their way back into the defaults.
* It should be possible to inherit from and modify the default configuration. 
    * This indicates we should have some ability to merge any/all configuration settings in a standardized way

## Format?

Providing a configuration file of a common format (explicitly not Python code) such as JSON or YAML seems to be superior for the following reasons:
* Easier to generate from code
* Easier to validate from code
* Less tied to API changes (but not entirely shielded, of course)
* Syntax errors or API changes won't break loading entirely

## Device classes and instances

Why support classes as well as instances?
- Requiring the user to copy/paste configuration for every instance of `EpicsMotor` would be nonsensical (and generate a large config file, at that)

Why not just support classes and not instances?
- Our goal in the end is to provide granular configuration for each and every device.
- It's impossible to confidently say "the user will never want different settings for two of the same device class". (We also haven't figured out exactly all of the configuration settings we'd like to support yet, at that)

## Inheritance?

Questions for now, as this requires more discussion:
- How does inheritance apply to configuration?
- In some cases, we may want to inherit and add to base class configuration
- In some cases, we may want to start from scratch
- Merging configuration from an inheritance tree could be messy.
[...]

PR #337, opened 2022-05-04
ENH: Add conf.yml key for loading and applying custom device tab completion whitelists

tangkong:
## Description
- adds `obj_config` key to VALID_KEYS
- adds `update_whitelist` process to be used in the `configure_objects` step
- integrates `configure_objects` step into `load_conf` procedure

image

## Motivation and Context
[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-123)

**I am very open to suggestions and critiques here**

In the request for custom whitelist modifications, it was suggested that there may other settings that might come up in the future.  This PR tries to pave the way for such configuration options.  

Currently `configure_objects` only calls `update_whitelist` and `update_blacklist`, but in the future other functions might also be called from that same point, reading the same yaml file that specifies the tab whitelists.  A helper function has been added to recurse through namespaces.

## How Has This Been Tested?
Interactively.  A test case has been added. 

This PR, docstrings, etc.  I wrote a new docs page for this too.


hutch_python/tests/test_user_load.py
```diff
@@ -14,3 +19,63 @@ def test_user_load():
...
+
+@pytest.mark.parametrize(
+    'objs', ['flat', 'nested'], indirect=True
+)
+def test_indirect(objs):
+    ns, dev_a, dev_b = objs
+    cfg = {'a': {'tab_whitelist': ['bar']},
+           'b': {'tab_whitelist': ['bar', 'dne']}}
+
+    assert 'bar' not in dev_a.__dir__()
```
klauer, line 51:
```suggestion
    assert 'bar' not in dir(dev_a)
```

klauer:
I like the implementation overall. It's pretty simple/friendly compared to what I had been envisioning - a very good thing.
As-is, this allows for "add item to tab list" and "remove item from tab list"
Do you think there's room for a key that says "only show these attrs in tab completion and none that are defined in pcdsdevices?".

hutch_python/tests/test_user_load.py
tangkong: I've become too comfortable with 'dunder' methods...

klauer:
Another thought: how about persisting tab completion information from the hutch-python prompt back into the config file?

tangkong:
You are correct.  

I think it should be simple enough to add a "replace tab list".  I'd love to hear of any non-tab-completion modifications to objects that might be useful, since I was hoping  this would be able to cover that as well.  

My initial thought about persisting the information back to the config files was that it would take machinery similar to what positioner Presets uses.  All of that lives in `pcdsdevices`, so maybe a followup PR would be appropriate.  

My subsequent thoughts about writing the configuration are:  Since the config records a series of modifications (add/remove/replace), would we have to then record each change made to the device?  And then possibly distill those changes into the most concise set of instructions?  Seems interesting, though a bit involved for the tab stuff

klauer:
We had a whole set of discussions about this, after which point I dropped it on the floor: https://github.com/pcdshub/hutch-python/issues/293

tangkong:
The ci error we saw in zach's run-wrapper PR is back, though I thought we changed something upstream that fixed it.  (Zach committed, then reverted a `matplotlib=agg` setting in travis.yml).

ZLLentz:
Looks like this might be related to the pyqt=5.15 errors we've seen elsewhere

klauer:
Yep:
```
pyqt                      5.15.4           py39h5a03fae_0    conda-forge
```
Pin it and this should start passing again. Would be advisable to do that in a separate PR though.

hutch_python/user_load.py
```diff
@@ -50,3 +56,181 @@ def get_user_objs(load, *, ask_on_failure=True):
                 )
 
     return objs
+
+
+def update_objs(
```
klauer, line 61: Random question unrelated to this specific line:

Could I use the object configuration from my device class implementation (i.e., from the pcdsdevices side of things)? That is to say, if I want to store user-configurable data for MyDevice, would it fit into this scheme?

tangkong: If the question is: "can I use this configuration framework to set attributes?", I imagine the answer is yes.  We'd just have to write an `update_data` function that accesses the device right?

klauer: I wasn't sure if it was built in here already without re-reading the source - so I think that is my question, then.

If we all think it's a good thing to have, the next step would be: what would be a good interface to do that? A `BaseInterface` property that does something like `device.conf_yml_data = settings`? A method `set_conf_yml_data` (named differently or made semi-private)?

hutch_python/user_load.py
```diff
@@ -50,3 +56,181 @@ def get_user_objs(load, *, ask_on_failure=True):
...
+    dev = getattr(obj_ns, entry, None)
+    if dev:  # look for explicit device
+        fn(dev, attrs)
+
+        return
+
+    # look for devices with class
+    for dev in obj_ns:
+        if isinstance(dev, utils.HelpfulNamespace):
+            update_objs(dev, entry, attrs, fn)
```
ZLLentz, line 95: How do we get to this branch in a useful way? Do the nested namespaces have to have some collusion-level coordination on the entry string? I feel like my brain rolled off the code at some point

hutch_python/user_load.py
```diff
@@ -50,3 +56,181 @@ def get_user_objs(load, *, ask_on_failure=True):
...
+        # a poor-man's type check, in an effort to avoid importing
+        # every device type again
+        if entry in str(type(dev)):
```
ZLLentz, line 98: This is a bit worrysome for devices whose classes share similar substrings

hutch_python/user_load.py
```diff
@@ -50,3 +56,181 @@ def get_user_objs(load, *, ask_on_failure=True):
...
+    obj_ns is a HelpfulNamespace, which may include other
+    HelpfulNamespace's.  As a result this must be semi-recursive.
+
+    Parameters
+    ----------
+    obj_ns : HelpfulNamespace
+        Namespace of all objects
+
+    entry : str
+        Name of device to edit
```
ZLLentz, line 78: Oh, is this not just a name of a device, e.g. it can be a class name? Is that what I'm confused in my other comments?

hutch_python/user_load.py
```diff
@@ -50,3 +56,181 @@ def get_user_objs(load, *, ask_on_failure=True):
...
+    dev : BaseInterface
+        A single object implementing the tab completion features in
+        pcdsdevices.interface.BaseInterface
+
+    attrs : dict
+        configuration for kind settings on ``dev``
+    """
+    for cpt, kind in attrs.items():
+        k = getattr(Kind, kind, None)
+        if not k:
```
ZLLentz, line 175: Tiny nitpick corner:
- single variable names make me and my find and replace tools sad
- `not` checks are a little less precise when the only options are "an object" or "None". I'd always do these as `if kind is None` for clarity.

ZLLentz (commented):
I think I got really confused at some point reading this- brain might need to come back tomorrow. Overall I like.

hutch_python/user_load.py
klauer: We only work on a per-class basis here, correct? I think my configuration thing may be instance-level at times. Not sure how that could/should be built-in. (@ZLLentz any thoughts on this thread?)

ZLLentz: The idea you're discussing here is: an API that a class can follow to signify to hutch-python that a class/instance setting can be configurable via this config file, rather than needing to define the specifics externally in `hutch-python`? Is that correct?

First thought: we could support something like that with explicit entrypoint loading or duck-typed methods- hutch-python could see that an object or class has an unknown definition in the config file and check the class for how to handle it. This would probably work reasonably well. I think this second one is what you're suggesting a few comments above?

Second thought: for things that are truly just setattrs... could there be a default handler? A lot of class and object configurations don't need to involve running special code, just class.my_setting = 3

More thought needed maybe to consider the merits of different approaches here

docs/source/object_configuration.rst
```diff
@@ -0,0 +1,159 @@
+.. _object-configuration:
+
+====================
+Object Configuration
+====================
```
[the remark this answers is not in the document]
tangkong: Good points all around. 

> When to use this vs beamline.py?

Perhaps the most clear reason to do this is for the tab completion, where the API wasn't public.  Things like editing the device kind are easy enough to do in `beamline.py`, so I could see this being redundant.  

> fill this out easier, edit files through hutch-python

I was thinking about this briefly, and got stuck on how to remember the original state and isolate user changes.  If this was more attainable, I could see an interface similar to what we do with presets, where everything can happen through hp.

hutch_python/user_load.py
ZLLentz: I think the more precise way to do a soft no-import check is to check for equality against `type(dev).__name__` (hope I didn't mess up that syntax).

Do we care at all about inheritance in these? That would be the main reason to do real import loading.

hutch_python/user_load.py
tangkong: If I'm understanding the question right: the way we hit this recursive step is because I pass in the `all_objects` namespace, which also has HelpfulNamespaces inside it (`sim`, `m`, `x`, etc).  

Reading over your comment again I'm very not sure what the question is.

ZLLentz: I'm trying to remember what I was thinking about last week, I think I may have just misunderstood the code

hutch_python/obj_config.py
```diff
@@ -0,0 +1,189 @@
...
+        List of attributes to add to whitelist
+
+    fn : Callable[[BaseInterface, dict], None]
+        A function that applies a change to a singular device
+    """
+    dev = getattr(obj_ns, entry, None)
+    if dev:  # look for explicit device
+        fn(dev, attrs)
+
+        return
```
ZLLentz, line 43:
```suggestion
        fn(dev, attrs)
        return
```

hutch_python/obj_config.py
```diff
@@ -0,0 +1,189 @@
...
+    ----------
+    dev : BaseInterface
+        A single object implementing the tab completion features in
+        pcdsdevices.interface.BaseInterface
+
+    attrs : list[str]
+        list of attributes to replace ``dev`` tablist with
+    """
+    for old_att in dir(dev):
+        dev._tab.remove(old_att)
```
ZLLentz, line 106: I don't think this is exhaustive, since the tab completion mechanism lets us use regex as specifiers. It'd be easiest to just call reset I think: https://github.com/pcdshub/pcdsdevices/blob/4c01ba8892215c535e7e6ee2ac5a418f5aea4a07/pcdsdevices/interface.py#L71

hutch_python/obj_config.py
```diff
@@ -0,0 +1,189 @@
...
+    attrs : list[str]
+        list of attributes to add to ``dev`` tablist
+    """
+    for att in attrs:
+        dev._tab.add(att)
+
+
+def update_blacklist(dev: BaseInterface, attrs: list[str]) -> None:
+    """
+    Remove each field in ``attrs`` from tab whitelist for ``dev``
```
ZLLentz, line 74: This will fail in the niche case where something is in the tab whitelist by virtue of a regex and the user tries to blacklist it... not sure what to do about it...

hutch_python/obj_config.py
```diff
@@ -0,0 +1,189 @@
...
+def update_kind(dev: BaseInterface, attrs: dict) -> None:
+    """
+    Update the kind of components in ``dev``.  If a key matches the
+    device name, the top-level device kind will be modified.
+
+    Parameters
+    ----------
+    dev : BaseInterface
+        A single object implementing the tab completion features in
+        pcdsdevices.interface.BaseInterface
```
ZLLentz, line 121: A bit pedantic of me, but I think the description here is too narrow. This  kind updater should work on any ophyd device, right?

ZLLentz (commented):
A few more small things, overall I'm really happy with this and am happy to move on from the world where the scientists need to ask us to manage their tab completion preferences.

hutch_python/obj_config.py
tangkong: If you mean just for `update_kind`, you're absolutely right.

hutch_python/obj_config.py
tangkong: Are users adding to the regex manually?  It looks to me that the regex is cleared with `add` and `remove`, then rebuilt when `__dir__` is called.  

Are users supplying their own tab completion patterns, and is that the usage pattern we're targeting?  There's the `tab_whitelist` attribute, but adding to that doesn't actually change the whitelist?

tangkong: This is an issue in the tab completion interface in general, I think.  I can take a stab

hutch_python/user_load.py
klauer: I think there are a lot of paths forward here - and it probably doesn't need to be resolved here. Seems like I should open an issue to summarize our suggested approaches and see what we can do about fitting it into the current scheme (since we seem to agree it could be useful).

hutch_python/user_load.py
klauer: I agree that the general `if entry in name` needs refactoring - equality check seems appropriate.

I'm stuck between:
* Inheritance-level configuration makes it so easy to change settings for all devices consistently!
* Inheritance-level configuration is too confusing, will be troublesome for us to trace back settings with, and will be a pain for users

hutch_python/obj_config.py
```diff
@@ -0,0 +1,186 @@
...
+import yaml
+from ophyd import Device, Kind
+from pcdsdevices.interface import BaseInterface
+
+from . import utils
+
+logger = logging.getLogger(__name__)
+
+
+def update_objs(
```
ZLLentz, line 14: Does this function need to show warnings when we didn't find any matches?

Imagine I misspell an object or a class name and then my tab settings don't apply. Is this something we need to worry about happening?

tangkong: Definitely.  A good thing to add

hutch_python/obj_config.py
tangkong: I'm running into problems using `_tab.reset()`, as the failing CI attests to.  Phenomenologically, `reset` seems to just revert the tab settings to the default state, but looking at the code I'd expect it to clear out all the settings. There's some strange subclassing inheritance stuff going that I don't understand, maybe?

tangkong: I can read, and understand better now.  `_tab.remove` is better than `_tab.reset()`, but still doesn't catch regex examples.

hutch_python/user_load.py
tangkong: My thought was that typical hutch-python sessions have lots of similar devices (e.g. IMS motors) that might want similar configurations.  I agree that inheritance-level configuration might not be worthwhile, but "edit all devices of a specific class" might still have a place.

tangkong:
I think I'm going to argue further edge cases (removing items exposed by regex matches) are out of scope for this PR, and punt them to another PR.  

I'll open issues covering the thoughts covered in the conversations here.

ZLLentz (approved):

klauer (approved):
I'm really happy with what you've provided here. It's a great basis for this configuration functionality which will make our lives and our users' lives easier. 👍 

And it's even well-documented AND tested 💯

PR #340, opened 2022-05-19
ENH: subscribe BECOptionsPerRun instead of default BEC

tangkong:
## Description
Subscribes the modified BECOptionsPerRun, as defined in pcdshub/nabs#61 .  This provides the ability to specify BEC settings through plan metadata. 

## Motivation and Context
[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-1090)

Brings the changes pcdshub/nabs#61 to the people

## How Has This Been Tested?
Interactively

This PR.


hutch_python/load_conf.py
```diff
@@ -16,6 +16,8 @@
 from bluesky.callbacks.best_effort import BestEffortCallback
 from bluesky.callbacks.core import LiveTable
 from bluesky.callbacks.mpl_plotting import initialize_qt_teleporter
+from event_model import RunRouter
```
ZLLentz, line 19: I guess I'll finally look up what RunRouter is- code here looks good (the beef of the code is in the other PR)

ZLLentz:
This probably needs a nabs tag, and a minimum version pin, right?

tangkong:
I guess this CI will fail until we tag nabs?

ZLLentz:
Best practice would be to tag nabs and then re-run this, if we're trying to be good about this sort of procedural thing. Tags are easy and cheap though so it's probably ok.

PR #348, opened 2022-08-12
TST: address failing tests on psbuild

ZLLentz:
## Description

Some of the tests fail on psbuild in the latest conda environment, but not on travis CI. This PR will make the test suite run on our local servers, adjusting the test suite as appropriate.

- [x]  Mark `test_bug` as xfail (TODO: add issue for removing this in favor of the jira function from pcdsutils?)
- [x] Fix logging cross contimination bug that affected `test_set_console_level`, `test_debug_mode`, `test_debug_context`, and `test_debug_wrapper`
- [x] Fix race condition in test_sigquit_two_hits by cleaning up ophyd objects properly to have more stable thread execution
- [x] Fix phantom skips caused by a bad lightpath import (will probably need to update again after lightpath rework)
- [x] Fix log ophyd put spam in random tests by cleaning up ophyd properly

## Motivation and Context

We need to make sure these aren't real bugs. The automated suite missing failures is scary.
- `test_bug` tests a feature that has not been used in more than a year: posting to https://github.com/pcdshub/Bug-Reports-and-Requests/issues. It fails on recent versions of ipython when running on real computers. It passes on CI because it hits a caught exception first that we ignore because we're expecting it to happen on CI only. To fix this function we need to understand a bit better how the ipython session numbers work which is opaque (I tried for a bit). It's more prudent to move on and replace this with the jira function later. It'd be nice to revisit this so we can include ipython session context in the jira reported.
- The various logging configuration tests were failing in two ways: one when run individually due to logging shenanigans (not fixed, very confused) and another way when run in the full suite (logging cross-contamination). It wasn't too hard to fix the contamination by checking all the messages in the queue handler instead of only the most recent one.
- The race conditions seemed to be caused by the 120Hz epics updates running in the back that are logged to file boxing out other threads that needed to run. Cleaning up ophyd helps isolate the tests.
- The phantom skips seemed to have been caused by a typo or a missed refactor from long ago.

## How Has This Been Tested?

Run test suite locally

N/A

### DIFF
--- a/hutch_python/tests/conftest.py
+++ b/hutch_python/tests/conftest.py
@@ -13,6 +13,7 @@ import zmq
 from epics import PV
 from ophyd.areadetector.plugins import PluginBase
 from ophyd.device import Component as Cpt
+from ophyd.ophydobj import OphydObject
 from ophyd.signal import Signal
 from pcdsdevices.areadetector.detectors import PCDSAreaDetector
 
@@ -246,3 +247,40 @@ class DummyZMQSocket:
 def dummy_zmq_lcls2(monkeypatch):
     # monkeypatch such that the lcls2 daq doesn't actually connect to things
     monkeypatch.setattr(zmq, 'Context', DummyZMQContext)
+
+
+this_test_ophydobj = []
+
+
+@pytest.fixture(scope='session', autouse=True)
+def register_ophydobj():
+    """
+    Save references to all the ophydobj we create.
+    """
+    OphydObject.add_instantiation_callback(this_test_ophydobj.append)
+
+
+@pytest.fixture(scope='function', autouse=True)
+def cleanup_ophydobj():
+    """
+    Disable ophydobj that were created by the test.
+
+    We need to do this or they can persist between tests, clogging the logs
+    as the PVs update and causing race conditions.
+    """
+    yield
+    for obj in this_test_ophydobj:
+        # Should probably destroy here but it segfaults
+        # Need to remove all subs, disable the monitor, remove callbacks
+        # Might fail on non-pyepics idk
+        obj.unsubscribe_all()
+        for pv_attr in ('_read_pv', '_write_pv'):
+            try:
+                pv = getattr(obj, pv_attr)
+            except AttributeError:
+                pass
+            else:
+                pv.auto_monitor = False
+                pv.clear_callbacks()
+    this_test_ophydobj.clear()
+
--- a/hutch_python/tests/conftest.py
+++ b/hutch_python/tests/conftest.py
@@ -26,7 +26,7 @@ except ImportError:
     HutchELog = None
 try:
     import lightpath
-    from lightpath import beamlines
+    from lightpath.config import beamlines
 except ImportError:
     lightpath = None
     beamlines = None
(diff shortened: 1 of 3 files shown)


ZLLentz:
- `test_bug` can be skipped after some analysis and realizing it's kind of hard to fix (we don't actually use the function right now)
- the console tests fail in slightly different ways if they are run alone vs in the full suite. When run alone there is a complete failure I don't understand yet (no messages) and when run in the full suite it seems to be a race condition with the signal updates from other tests outliving their test bounds and getting picked up instead of the test message
- `test_sigquit_two_hits` is not something I expected to see here but it also failed once on psbuild, I need to run this a few dozen times to see how frequently it fails and in what way

I need to update the desc with the tests I'm expecting to fix

ZLLentz:
The console tests pass alone if you run them via `python run_tests.py` but fail if you run them via `pytest` (because no logging is configured?) or if you run them in the context of the full library (because of the logging cross-contamination?). Bizarre. I'm not sure why this isn't caught on CI but it's at least a test artifact rather than an outright bug.

ZLLentz:
It wasn't too hard to make the console tests robust to logging cross-contamination, but I'm dizzy trying to unravel the secrets of logging again that prevent the tests from passing when run individually. I'm going to overlook that for now, it can become an issue for later.

ZLLentz:
Luckily, only the bug report was a real bug and it doesn't seem to be in use.

hutch_python/tests/conftest.py
```diff
@@ -246,3 +247,40 @@ def recv_json(self, *args, **kwargs):
...
+@pytest.fixture(scope='session', autouse=True)
+def register_ophydobj():
+    """
+    Save references to all the ophydobj we create.
+    """
+    OphydObject.add_instantiation_callback(this_test_ophydobj.append)
+
+
+@pytest.fixture(scope='function', autouse=True)
+def cleanup_ophydobj():
```
klauer, line 264: These would be great candidates for something like `pcdsutils.tests.ophyd_helpers`

klauer (approved):

hutch_python/tests/test_log_setup.py
```diff
@@ -84,16 +85,20 @@ def clear(queue):
 
 def assert_is_info(queue):
     clear(queue)
-    logger.info('hello')
-    logger.debug('goodbye')
-    assert 'hello' in queue.get(block=False).getMessage()
-    assert queue.empty()
+    info_msg = str(uuid.uuid4())
+    debug_msg = str(uuid.uuid4())
```
tangkong, line 89: This was changed to avoid logging cross contamination?

tangkong (approved):
LGTM

ZLLentz:
My mistake for failing to have pre-commit running on this repo for me. That's the travis failure.

hutch_python/tests/conftest.py
ZLLentz: Added an issue

hutch_python/tests/test_log_setup.py
ZLLentz: I thought it would be better to use random garbles of strings rather than messages that could theoretically be matched now that I'm searching through potentially several messages instead of exactly one.

PR #356, opened 2023-02-11
FIX: ensure endstation devices are always loaded

ZLLentz:
## Description

Add the endstation name to the beamlines to load set.

## Motivation and Context

It's possible for this to be missing if the hutch's own beamline isn't in its lightpath.
This is... rare? Or maybe it only happens due to a misconfiguration?
Whatever the case, it's an edge case to cover.

In most cases, this won't do anything because the hutch name should already be in the set.

## How Has This Been Tested?

Interactively only via getting cxi devices to load again.

Release notes only

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ exclude: |
     )$
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
     hooks:
     -   id: no-commit-to-branch
     -   id: trailing-whitespace
@@ -28,11 +28,11 @@ repos:
     -   id: debug-statements
 
 -   repo: https://github.com/pycqa/flake8.git
-    rev: 3.8.3
+    rev: 6.0.0
     hooks:
     -   id: flake8
 
 -   repo: https://github.com/timothycrosley/isort
-    rev: 5.6.4
+    rev: 5.12.0
     hooks:
     -   id: isort
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,13 @@ env:
     # Extra dependencies needed to run the tests which are not included
     # at the recipe and dev-requirements.txt. E.g. PyQt
     # pandoc/pypandoc - used to translate README.md for sphinx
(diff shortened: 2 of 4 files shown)


tangkong (approved):
Looks good to me.  Ensures any device with beamline=<hutch_name> will get loaded, even if not lightpath-valid.

klauer (approved):

ZLLentz:
Short-term it is OK to pin `ipython` in a repo like `hutch-python` which has no offsite usage.
It might be most convenient to pin it in the CI only so that when we go to test newer versions we don't have to fight the conda environment solver.

ZLLentz:
I think I should probably do the pinning during this PR to show working CI
I'll use the ipython that we run on the beamlines: `ipython=8.4.0`

This is still using Travis but maybe it's ok to make the adjustment now and keep it in mind when we get around to gha-ifying this one

klauer:
The GHA migration tools _should_ pick up the pin from the Travis configuration.
This looks good to me, except: https://app.travis-ci.com/github/pcdshub/hutch-python/jobs/595983014
We need a `pre-commit autoupdate` here to fix that traceback

tangkong:
I suppose pinning Ipython in the test suite alone makes sense, since we don't interact with the deprecated portion of the API in the code base itself.  

A perhaps naive question:
When we build the environment, we run the each package's test suite.  What happens in this case?  Will it take this pin and pass tests, being different from the built environment?  Or will the environment use a recent ipython and fail the test suite?

ZLLentz:
The main motivation for pinning in this way is to avoid a situation where I'm trying to test new versions of ipython and having conda say "no! you aren't allowed to use that ipython version!" at the install level. = pins and  < pins cause a lot of issues that aren't solvable without modifying the recipe, but shaky >= pins are simple to work around.

ZLLentz:
> When we build the environment, we run the each package's test suite. What happens in this case?

The environment build would happily using the new ipython if it wasn't pinned, but it is pinned at the environment level.

ZLLentz:
I haven't investigated far enough yet to build an understanding if this is an incompatibility for usage or if it's just a testing artifact

Issue #357, opened 2023-02-15
BLD: test suite fails with latest ipython


ZLLentz:
## Expected Behavior

tests should pass

## Current Behavior

with the latest ipython, there have been enough api or functional changes that the test suite no longer passes

## Possible Solution

- investigate what these tests represent and if the failure represents a real issue or a testing issue
- revise the package and/or tests as appropriate for compatibility

## Context

#356

tangkong:
From my brief investigations, `hutch-python` sets a couple of attributes on the ipython `Completer`, and aside from that doesn't interact with it much.
https://github.com/pcdshub/hutch-python/blob/954d4cc616528856a189cd813797162ea8fe90a6/hutch_python/cli.py#L46-L70

The test suite dove deeper into the internals (?), hitting upon an attribute that got deprecated.  The `line_buffer` attribute is not guaranteed to exist [(gh)](https://github.com/ipython/ipython/blob/2a5fdf9dab8e5fa50a03ce686aaf11cb38844587/IPython/core/completer.py#L3100), so calling the `attr_matches` ([gh](https://github.com/ipython/ipython/blob/2a5fdf9dab8e5fa50a03ce686aaf11cb38844587/IPython/core/completer.py#L1142)) independently can lead to the attribute error as observed.

I'm not sure what ipython wants the public API to be here, it's not been annotated very clearly from the code itself.

tangkong:
Reviving this in an effort to move to py3.12, which requires more updated ipython versions.  We currently use ipython 8.4.0, and though IPython does not explicitly state when it became py3.12 compatible, it's at least[ after 8.14](https://ipython.readthedocs.io/en/stable/whatsnew/version8.html#python-3-12-compatibility).  Py3.12 bugs were being addressed as [late as 8.32](https://ipython.readthedocs.io/en/stable/whatsnew/version8.html#ipython-8-32), though CI was being [run on py3.12 as of 8.19](https://ipython.readthedocs.io/en/stable/whatsnew/version8.html#remove-of-python-3-9-support)

Can we upgrade further?  I'd be afraid if I'm honest.  Ipython 9.x involved a significant refactor and removal of a number of deprecations.  But it adds an LLM suggestion hook so we need it right now!

PR #361, opened 2023-04-04
DOC: run release notes script for v1.18.1

ZLLentz:
## Description

Get release notes ready for the tag

## Motivation and Context

I want to tag every package that I help maintain since we haven't done this at all in 2023.

## How Has This Been Tested?

n/a

is docs


tangkong (approved):
Seems like we've been ignoring the tests that fail here, or at least procrastinating on fixing them.  This should be good to go probably.

ZLLentz:
You're right, looks like I've put it on the backburner. See #357

ZLLentz:
The weird pip failure was only in a workflow running on my fork, rather than the workflow running here :confused:

klauer (approved):

ZLLentz:
Yes, IPython is currently on the "do not update this basically ever" list due to past pain with unexpected breaks or behavior changes between minor releases.

PR #363, opened 2023-04-10
TST: fix an issue with the test suite failing to run

ZLLentz:
## Description

- The test suite checks the version of `lightpath` using a function with strict type checking, so we need to coerce our UserString into a str.
- Add a release note with the intention of tagging this tomorrow

## Motivation and Context

Without this change, the test suite cannot be run in an environment with the latest `lightpath`.

## How Has This Been Tested?

Using the test suite

Release note

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -1,6 +1,19 @@
 Release History
 ###############
 
+v1.18.2 (2023-04-11)
+====================
+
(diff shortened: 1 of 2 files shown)


klauer (approved):
Makes sense and the fix looks good 👍

tangkong (approved):

PR #370, opened 2023-07-25
MAINT: unpin pyqt and fix test suite, REL: v1.18.4

### LINKED ISSUE #354: CI: ipython test broken again
Probably it's time to remove or replace this test case with something simpler and more direct

_Originally posted by @ZLLentz in https://github.com/pcdshub/hutch-python/issues/353#issuecomment-1316089563_

### LINKED ISSUE #362: docs-versions-menu missing in docs-requirements and Sphinx conf.py
https://github.com/pcdshub/hutch-python/blob/791755703867969c49805cf99e151ba444cfae8a/docs/source/conf.py#L36

Versions menu not showing up in docs: https://pcdshub.github.io/hutch-python/

### LINKED ISSUE #364: GHA docs fail to build for hutch-python
```
Running Sphinx v6.1.3
making output directory... done
[autosummary] generating autosummary for: beamline.rst, bug.rst, cache.rst, cli.rst, daq.rst, database.rst, debug.rst, environment.rst, experiment.rst, hutch_setup.rst, ..., report.rst, start.rst, startup.rst, tips.rst, tutorial.ipynb, upcoming_release_notes/template-full.rst, upcoming_release_notes/template-short.rst, user_utils.rst, utils.rst, yaml_files.rst

Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x7ff0cf91e5e0> for event 'builder-inited' threw an exception (exception: no module named hutch_python.load_conf)
Failed to import hutch_python.load_conf.
Possible hints:
* ModuleNotFoundError: No module named 'psdaq'
* AttributeError: module 'hutch_python' has no attribute 'load_conf'
Failed to import hutch_python.cli.
Possible hints:
* AttributeError: module 'hutch_python' has no attribute 'cli'
* ModuleNotFoundError: No module named 'psdaq'
make: *** [Makefile:[24](https://github.com/pcdshub/hutch-python/actions/runs/4670711097/jobs/8270782139#step:15:25): html] Error 2
Error: Process completed with exit code 2.
```

ZLLentz:
## Description

- pyqt doesn't need to be held back any more, it just needs to be some flavor of pyqt5
- fix a bug where non-conda installs and non-standard conda installs (like mambabuild intermediates) would be unable to use the `hutch-python --create` cookiecutter stuff
- add send and recv methods to the fake zmq socket because newer versions of the lcls2 daq code are expecting recv
- fix pyproject.toml typo docs -> doc
- update requirements.txt for accuracy now that psdaq-control-minimal is available on pypi
- include docs-versions-menu

This also adds release notes for v1.18.4

## Motivation and Context

Allows us to build py311 environments, probably

closes #364 
closes #362
closes #354


tangkong (approved):
free the qt! ✊

ZLLentz:
Wow what a wild CI wipe, need to look closer at it

ZLLentz:
A few things here:

- we'll not be able to see the full affects of this until tagging nabs and hxrsnd
- there's some incomprehensible error in the conda builds
- I keep messing up the pypi requirements sytax

ZLLentz:
After the other tags I'll try again to reproduce some of these offline

klauer:
Cookiecutter failure (`hutch-python --create`) appears to be hitting this code path:
https://github.com/pcdshub/hutch-python/blob/4a58e063def389c35f18c68098e6288ff4c8e0b1/hutch_python/cli.py#L161-L162
Resulting in a `pathlib.Path` getting into the jinja template rendering

ZLLentz:
For the record, I managed to create an exact copy of the conda env and could not reproduce any of these issues. I'll try again later.

When run locally, there are some shenanigans though: the cookiecutter test does entirely the wrong thing and still passes (it doesn't go through the cookie cutter code at all, it loads a beamline session instead). I'll need to wrap my head around what's going on here.

ZLLentz:
I assumed that the pytest logging cli options were related to the logging issues but it seems like it's not true. The next thing to do is to try from a fresh clone in case some of the local generated files cause it to pass.

klauer:
> I can confirm that the ipython version is held back in the ci settings right now

conda job:
`ipython                   8.14.0             pyh41d4057_0    conda-forge`

pip job:
`ipython                       8.4.0`

ZLLentz:
Look more closely, the `conda list` is done prior to the final install step

ZLLentz:
I'm not 100% sure, but I think the failing create test was also failing to restore the logging config properly. The number of failing tests has dropped to just 1: the ipython banner check test.

ZLLentz:
I do see the above locally, but it doesn't cause a test failure locally. Should I look more closely at it?

The test that fails on pypi is actually a bugged test- it will never pass, since it checks for the IPython banner that we've replaced. The only reason the conda build (and the local build) passes is that this test is incidentally skipped.

klauer:
I wrote the above comment wondering about the failures, then noticed tests were passing magically, then saw your comment... So, no, I think the above warnings are probably not worth further investigation

ZLLentz:
Just a bit more cleanup left to do, sphinx_rtd_theme isn't in the docs build so I probably have to fix the pyproject.toml

And then I guess I've done enough that I should edit the PR text and pre-emptively include a release notes docs page

ZLLentz:
If the docs that this final push generates look OK then this should be done. There is more to do but this is far enough for today.

klauer (approved):
LGTM 👍

ZLLentz:
The docs built correctly, so away we go!

PR #376, opened 2023-10-06
ENH: Adding --update feature to epicsarch-qs (in progress)

c-tsoi:
## Description

I am currently working on adding an --update feature that looks at the current hutches epicsArch file and compares it to the questionnaire. It looks for instances of duplicate PVs and if this occurs then it will take the alias from the questionnaire instance and replace it for the one in the local epicsArch file.
## Motivation and Context

This change helps to prevent duplicating PV entries across the arch files, ensures we have the most current information from the questionnaire and gives us another option besides overwriting the current file.
## How Has This Been Tested?

Untested/Still in progress. (Going on vacation and want to update progress)

N/A

## Pre-merge checklist

### DIFF
--- a/hutch_python/epics_arch.py
+++ b/hutch_python/epics_arch.py
@@ -7,7 +7,8 @@ import warnings
 
 from collections import OrderedDict
 from .constants import EPICS_ARCH_FILE_PATH
-from .qs_load import get_qs_client
+from .qs_load import get_qs_client, pull_cds_items
+#from .qs_load import get_qs_obj
 from itertools import chain
 logger = logging.getLogger(__name__)
 logging.basicConfig(level="INFO")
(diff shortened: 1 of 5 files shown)


hutch_python/epics_arch.py
```diff
@@ -28,19 +28,25 @@ def _create_parser():
 
     parser.add_argument('--dry-run', action='store_true', default=False,
                         help='Print to stdout what would be written in the '
-                        'archFIle.')
+                        'archFile.')
+
+    parser.add_argument('--update', action='store_true', default=False,
```
vespos, line 33: I dont think we need that argument, it serves the same purpose as --dry-run (in fact it does the exact opposite).

c-tsoi: Sounds good! I am working on this now

PR #378, opened 2024-03-21
Epicsarch qs

c-tsoi:
## Description

I have added several feature edits and a couple of new features. Changes are listed below:

--update: feature that looks at the current hutches epicsArch file and compares it to the questionnaire. It looks for instances of duplicate PVs and if this occurs then it will take the alias from the questionnaire instance and replace it for the one in the local epicsArch file. If there is no existing archfile it will do the existing task of of pulling the questionnaire data into a fresh archfile.

--softlink: this feature gives the users to unlink the symbolic link to epicsArch_experiment_specific.txt. It then replaces that link to point to the archfile to the new experiment.

--level: this feature gives developers the ability to set a debugging level. This shows where the code faults in the case that it does.

--cds-items: this feature interacts with the Python Questionnaire Client, for a given run and experiment. It pulls all of the elements in the cds tab of the questionnaire in the form of a dictionary. From there is parses the dictionary for keys with the string "pcds-setup". It then sorts the keys by type and shows them in a PrettyTable output in the terminal.

I have also edited some of the test scripts to match the new method names. I have also skipped one of the test methods because the test is no longer representing the functionality of the method and interactive testing has been done on it.

## Motivation and Context

This change helps to prevent duplicating PV entries across the arch files, ensures we have the most current information from the questionnaire, and gives us another option besides overwriting the current file. It also provides helper functions for tasks and provides developer tools for debugging in the case of errors.

## How Has This Been Tested?

The majority of the testing has been done with the existing test scripts on the hutch python folder and for the methods that no longer help I have done some interactive testing.

(https://confluence.slac.stanford.edu/display/PCDS/Questionnaire+flight+rules)
This page is a WIP but will have descriptions and examples of how to use the new methods.

## Pre-merge checklist

### DIFF
--- a/hutch_python/epics_arch.py
+++ b/hutch_python/epics_arch.py
@@ -7,7 +7,8 @@ import warnings
 
 from collections import OrderedDict
 from .constants import EPICS_ARCH_FILE_PATH
-from .qs_load import get_qs_client
+from .qs_load import get_qs_client, pull_cds_items
+#from .qs_load import get_qs_obj
 from itertools import chain
 logger = logging.getLogger(__name__)
 logging.basicConfig(level="INFO")
@@ -33,6 +34,9 @@ def _create_parser():
                         'archFile.')
     parser.add_argument('--level', '-l',required=False, type=str, default="INFO",
                         help='Show the debug logging stream')
+    parser.add_argument('--cds-items', nargs=2, action='store',default=None, 
+                        help="Pulls all data from CDS tab. E.g.: xppx1003221 run21 X-10032")
+    parser.add_argument('--link', '-sl', action='store_true', default=None, help="create softlink for experiement")
     return parser
 
 
@@ -52,7 +56,7 @@ def cli_setup(args):
     logger.debug("Logger Level: ",logger.getEffectiveLevel())
     logger.debug("Set logging level of %r to %r", shown_logger.name, args.level)
 
-def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=False, update=False):
+def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=False, update=False, cds_items=None, link=None):
     """
     Create an epicsArch file for the experiment.
 
@@ -107,11 +111,24 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
             file_path = path
         elif hutch:
             file_path = EPICS_ARCH_FILE_PATH.format(hutch.lower())
+        elif cds_items:
+            pull_cds_data(experiment, cds_items)
+            return
+        elif link:
+            create_softlink(experiment)
+            return
         else:
             file_path = EPICS_ARCH_FILE_PATH.format(experiment[0:3])
         update_file(exp_name=experiment, path=file_path)
     elif dry_run:
         print_dry_run(experiment)
+
+def pull_cds_data(exp, run):
+    print("in client")
+    pull_cds_items(exp, run)
+    
+def create_softlink(exp):
+    print("in softlink")
 
 def check_for_duplicates(qs_data, af_data):
 
@@ -153,7 +170,7 @@ def check_for_duplicates(qs_data, af_data):
     qsDict = {k:v.replace("\n", "") for k,v in qsDict.items()}
     sorted_qsDict = dict(sorted(qsDict.items()))
 
-
+    # If the archfile is not empty then clean it if not ,skip
     if len(af_data) > 0:
         # ArchFile
         afDict = dict(zip(af_data[::2], af_data[1::2]))
@@ -320,11 +337,11 @@ def update_file(exp_name, path):
     logger.debug("UpdateFile: qs_data:\n" + str(qs_data))
     
     logger.debug("\nPath: "+ str(path))
-    af_path = path + "epicsArch_" + exp_name + ".txt"
+    af_path = str(path) + "epicsArch_" + str(exp_name) + ".txt"
     logger.debug("\nAF Path: " + str(af_path))
     exp_name = str(exp_name)
-    file_path = ''.join((path, 'epicsArch_', exp_name, '.txt'))
-    if not os.path.exists(path):
+    file_path = ''.join((str(path), 'epicsArch_', str(exp_name), '.txt'))
+    if not os.path.exists(str(path)):
         raise OSError('Invalid path: %s' % path)
     # if the path exists but archfile does not, create af and pull qsd
     elif os.path.exists(path) and not os.path.exists(af_path):
@@ -338,6 +355,7 @@ def update_file(exp_name, path):
         af_data = read_archfile(af_path)
         cleaned_data = check_for_duplicates(qs_data, af_data)
 
+    # Write updates to the corresponding file
     with open(file_path, 'w') as f:
             for data in cleaned_data:
                 try:
--- a/hutch_python/epics_arch.py
+++ b/hutch_python/epics_arch.py
@@ -267,6 +267,7 @@ def print_dry_run(exp_name):
     """
     Updating experiment file.
     """
+
     af_path = EPICS_ARCH_FILE_PATH.format(exp_name[0:3]) + 'epicsArch_' + exp_name + '.txt'
     af_data = read_archfile(af_path)
     if not os.path.exists(af_path):
--- a/hutch_python/epics_arch.py
+++ b/hutch_python/epics_arch.py
@@ -3,11 +3,12 @@ import argparse
 import logging
 import os
 import sys
+import warnings
 
 from collections import OrderedDict
 from .constants import EPICS_ARCH_FILE_PATH
 from .qs_load import get_qs_client
-
+from itertools import chain
 logger = logging.getLogger(__name__)
 
 
@@ -40,7 +41,7 @@ def _create_parser():
 
 def main():
     """Entry point."""
-    print("\nepicsarch-qs test script")
+    print("\nepicsarch-qs test script, git")
     parser = _create_parser()
     parsed_args = parser.parse_args()
     kwargs = vars(parsed_args)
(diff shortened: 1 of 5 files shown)


hutch_python/epics_arch.py
```diff
@@ -37,10 +47,21 @@ def main():
...
 
 
-def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
+def logger_setup(args):
+    # Setting up the logger, to show the level when enabled
+    logging.getLogger().addHandler(logging.NullHandler())
+    shown_logger = logging.getLogger('epicsarch-qs')
+    logger.setLevel(args.level)
+    logger.debug("Logger Level: ", logger.getEffectiveLevel())
+    logger.debug("Set logging level of %r to %r", shown_logger.name, args.level)
```
ZLLentz, line 61: This logger setup is a bit confusing to me.
`shown_logger` is created and then not used or modified. Then, we set up `logger`. Then we use `logger` to falsely report that we've set the log level of `shown_logger`.

hutch_python/epics_arch.py
```diff
@@ -56,6 +77,8 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
     dry_run : bool
         To indicate if only print to stdout the data that would be stored
         in the epicsArch file and not create the file.
+    update : bool
```
ZLLentz, line 80: DOC/MAINT: `update` isn't used in the function body here, instead we are currently always updating regardless of input. `cds_items` and `link` are both used, but neither is documented up here.

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
             file_path = EPICS_ARCH_FILE_PATH.format(hutch.lower())
+        elif cds_items:
+            pull_cds_data(experiment, cds_items)
+            return
+        elif link:
+            update_file(exp_name=experiment, path=EPICS_ARCH_FILE_PATH.format(experiment[0:3]))
+            create_softlink(experiment)
+            return
         else:
             file_path = EPICS_ARCH_FILE_PATH.format(experiment[0:3])
```
ZLLentz, line 127: The sequencing in this code block is problematic.
- `cds_items` and `link` will not be checked if `path` or `hutch` is provided, since this is a big if/elif tree.
- In general, this if/elif tree originally was used to determine the filepath to use, so it's not a very convenient place to check unrelated arguments
- the `link` argument ignores all pre-existing `EPICS_ARCH_FILE_PATH` handling and does its own thing here
- it's not clear from the documentation that "cds_items" is not supposed to update the file and "link" is still supposed to update the file

You probably want to check `cds_items` in a separate if block, prior to checking for path, and you probably want to check `link` after the file is updated in the normal space below (to simply add the softlink step). I think the help strings also need to be more clear that `--link` still runs update but `--cds-items` does not.

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
-        create_file(exp_name=experiment, path=file_path)
+        update_file(exp_name=experiment, path=file_path)
     elif dry_run:
         print_dry_run(experiment)
 
 
+def pull_cds_data(exp, run):
+    logger.debug("in client")
+    pull_cds_items(exp, run)
```
ZLLentz, line 135: Nitpick: I don't think this function is very useful and it makes it slightly harder to navigate the source code

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+
+
+def create_softlink(experiment):
+    logger.debug("in softlink")
+    # remove the old soft link and add a new one (update), *THIS HAS NOT BEEN TESTED YET*
+    # this removes the softlink in the /cds/group/pcds/dist/pds/{}/misc/
+    # rm_result = subprocess.run(['unlink', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment[0:3].upper() + '_exp_specific.txt'])
+
+    # This adds a new softlink in /cds/group/pcds/dist/pds/{}/misc/
```
ZLLentz, line 144: DOC: you should probably correct the comments here and remake them into a docstring

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    subprocess.run(['ln', '-sf', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment + '.txt', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment[0:3].upper() + '_exp_specific.txt'])
```
ZLLentz, line 145: TST/MAINT: This is hardcoded to be from `EPICS_ARCH_FILE_PATH`, making it cumbersome to test offline and incompatible with the path arguments. Maybe this needs to allow custom supplied user paths (defaulting to the normal path, of course)

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    qs_data : list
+    af_data : list
+
+    Examples
+    --------
+    >>> epicsarch-qs xpplv6818 --dry-run
+
+    Returns
+    -------
+    updatea_arch_list : list
```
ZLLentz, line 165: typo
```suggestion
    updated_arch_list : list
```

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    # remove the old soft link and add a new one (update), *THIS HAS NOT BEEN TESTED YET*
+    # this removes the softlink in the /cds/group/pcds/dist/pds/{}/misc/
+    # rm_result = subprocess.run(['unlink', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment[0:3].upper() + '_exp_specific.txt'])
+
+    # This adds a new softlink in /cds/group/pcds/dist/pds/{}/misc/
+    subprocess.run(['ln', '-sf', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment + '.txt', EPICS_ARCH_FILE_PATH.format(experiment[0:3]) + 'epicsArch_' + experiment[0:3].upper() + '_exp_specific.txt'])
+
+
+def check_for_duplicates(qs_data, af_data):
+
```
ZLLentz, line 149: nitpick
```suggestion
```

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    Examples
+    --------
+    >>> epicsarch-qs xpplv6818 --dry-run
+
+    Returns
+    -------
+    updatea_arch_list : list
+        Updated list containing sorted alias, PVs.
+
```
ZLLentz, line 167: nitpick
```suggestion
```

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    # Questionnaire Data, removing whitespaces and newline chars
+    qsDict = dict(zip(qs_data[::2], qs_data[1::2]))
+    qsDict = {k.replace(" ", ""): v for k, v in qsDict.items()}
+    qsDict = {k.replace("\n", ""): v for k, v in qsDict.items()}
+    qsDict = {k: v.replace(" ", "") for k, v in qsDict.items()}
+    qsDict = {k: v.replace("\n", "") for k, v in qsDict.items()}
+    sorted_qsDict = dict(sorted(qsDict.items()))
+
+    # If the archfile is not empty then clean it if not ,skip
+    if len(af_data) > 0:
```
ZLLentz, line 188: Nitpick: A consequence here is that `afDict` and `sorted_afDict` will not be defined if there is no archfile data. This could cause issues later in the function. I suggesting adding an `else` clause where you set these to empty dictionaries.

edit: I see later that you return early in these cases, so I'm downgrading this to nitpick

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+        rev_keyDict.setdefault(value, list()).append(key)
+
+    pvDuplicate = [key for key, values in rev_keyDict.items() if len(values) > 1]
+    # Looking for duplicates of PVs in the questionaire
+    # also print out the alias for PV, change removing to warning operater to remove dup then rerun
+    for dup in pvDuplicate:
+        logger.debug("!Duplicate PV in questionnaire!:" + str(dup))
+        for value in rev_keyDict[dup][1:]:
+            logger.debug("Found PV duplicate(s) from questionnaire: " + value + ", " + sorted_qsDict[value])
+        raise Exception("Please remove duplicates and re-run script!")
```
ZLLentz, line 212: Nitpick: we can make this a more precise error class, such as ValueError or RuntimeError

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+
+    # Check to see if the archfile has any data in it
+    if len(af_data) == 0:
+        logger.debug("CFD: Case: no archfile given, returning cleaned questionnaire data\n")
```
ZLLentz, line 216: Nitpick: it isn't necessary to end log messages with explicit newlines

hutch_python/epics_arch.py
```diff
@@ -93,13 +116,138 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+    return updated_arch_list
+
+
+def read_archfile(exp_path):
+    if os.path.exists(exp_path):
+        with open(exp_path, "r") as experiment:
+            lines = experiment.readlines()
+        return lines
+    else:
+        raise OSError('ArchFile not found: %s' % exp_path)
```
ZLLentz, line 248: Q: What are we guarding against by raising our own OSError instead of using the error raised by open?

hutch_python/epics_arch.py
```diff
@@ -113,9 +261,30 @@ def print_dry_run(exp_name):
...
+        for item in updated_archFile:
+            print(item)
+
+
+def get_key(val, my_dict):
+    for k, v in my_dict.items():
+        if val == v:
+            return k
+    strError = ""
+    return strError
```
ZLLentz, line 287: Q: What's this blank error message used for?

hutch_python/epics_arch.py
```diff
@@ -182,16 +351,32 @@ def create_file(exp_name, path):
...
-    file_path = ''.join((path, 'epicsArch_', exp_name, '.txt'))
+    qs_data = get_questionnaire_data(exp_name)
 
-    logger.info('Creating epicsArch file for experiment: %s', exp_name)
+    logger.debug("UpdateFile: qs_data:\n" + str(qs_data))
 
+    logger.debug("\nPath: " + str(path))
+    af_path = str(path) + "epicsArch_" + str(exp_name) + ".txt"
+    logger.debug("\nAF Path: " + str(af_path))
+    exp_name = str(exp_name)
```
ZLLentz, line 361: nitpick: you seem to be repeatedly doing `str(exp_name)` both before and after this line. You can just do this once earlier and not need to repeatedly recast as `str` later. You can similarly do this for `path`.

hutch_python/qs_load.py
```diff
@@ -16,6 +26,90 @@
...
+    alias: str
+    pvbase: str
+    pvtype: str
+
+
+def pull_cds_items(exp, run):
+    """
+    Gather all user obejcts from the CDS tab in the questionnaire.
+    Parse objects and sperate them based on type.
+    Display them in the console vie PrettyTable.
```
ZLLentz, line 41:
```suggestion
    Gather all user objects from the CDS tab in the questionnaire.
    Parse objects and separate them based on type.
    Display them in the console via PrettyTable.
```

hutch_python/qs_load.py
```diff
@@ -16,6 +26,90 @@
...
+
+    Parameters
+    ----------
+    exp: ``str``
+        The experiment's name e.g. xppx1003221
+    run: ''str''
+        The run number e.g. run21
+    ----------
```
ZLLentz, line 49: DOC: I think these extra ------- lines will mess with the autodocs, but I'm not 100% sure
See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

ZLLentz (changes requested):
Hi, the core here is good. I pointed out a lot documentation issues (DOC) and typos, and a small number of important issues. I also had a some things that confused me (may or may not be issues) and a bunch of nitpicks.

hutch_python/epics_arch.py
```diff
@@ -28,7 +30,15 @@ def _create_parser():
 
     parser.add_argument('--dry-run', action='store_true', default=False,
                         help='Print to stdout what would be written in the '
-                        'archFIle.')
+                        'archFile.')
```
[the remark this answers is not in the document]
c-tsoi: Ah, yeah. The script should just run update_archfile when it is trying to update or create one. I can change the description to be more accurate.

hutch_python/epics_arch.py
c-tsoi: If I am not wrong, an empty string is the same thing as returning false or 0? Might be totally wrong though.

hutch_python/epics_arch.py
```diff
@@ -37,10 +47,21 @@ def main():
     parser = _create_parser()
     parsed_args = parser.parse_args()
     kwargs = vars(parsed_args)
+    logger_setup(parsed_args)
+    logger.debug("\nepicsarch-qs test script, git")
```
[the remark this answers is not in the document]
c-tsoi: This was just my way of knowing my version version of the script was loading before I knew about $ which epicsarch-qs

hutch_python/epics_arch.py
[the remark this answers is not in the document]
c-tsoi: Yeah, at first I was planning on integrating that script into this one. Must have forgotten to remove the old remnants.

hutch_python/epics_arch.py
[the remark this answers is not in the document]
c-tsoi: Yeah, at first I was planning on integrating that script into this one. Must have forgotten to remove the old remnants.

hutch_python/epics_arch.py
c-tsoi: Will change to return None

hutch_python/epics_arch.py
```diff
@@ -28,7 +30,15 @@ def _create_parser():
...
     parser.add_argument('--dry-run', action='store_true', default=False,
                         help='Print to stdout what would be written in the '
-                        'archFIle.')
+                        'archFile.')
+
+    parser.add_argument('--level', '-l', required=False, type=str, default="INFO",
+                        help='Show the debug logging stream')
+
+    parser.add_argument('--cds-items', nargs=2, action='store', default=None,
+                        help="Pulls all data from CDS tab. E.g.: xppx1003221 run21 X-10032")
```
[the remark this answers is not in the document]
c-tsoi: Spoke with you about this, names of runs are too inconsistent.

hutch_python/epics_arch.py
ZLLentz: Another side thought I had offline earlier: another approach for these cases where args aren't respected is to simply error out when someone gives an incompatible arg. E.g. you can raise an exception or exit early when someone tries to link while also having a custom path.

c-tsoi: Yeah, I can probably hand this with an if path exists os thing.

hutch_python/epics_arch.py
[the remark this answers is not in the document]
ZLLentz: Re-marking as unresolved because I still think the help next needs to be clarified

hutch_python/epics_arch.py
c-tsoi: I have added another argument to parse args that supports custom file paths and it will error out in the case of an invalid path.

hutch_python/epics_arch.py
```diff
@@ -95,11 +140,204 @@ def create_arch_file(experiment, hutch=None, path=None, dry_run=False):
...
+        elif re.match('pcdssetup-trig.*-name', k):
+            pv = cds_dict.get(re.sub('name', 'pvbase', k), '')
+            displayList.append(QStruct(v, pv, "triggers"))
+        elif re.match('pcdssetup-vacuum.*-name', k):
+            pv = cds_dict.get(re.sub('name', 'pvbase', k), '')
+            displayList.append(QStruct(v, pv, "vacuum"))
+        elif re.match('pcdssetup-temp.*-name', k):
+            pv = cds_dict.get(re.sub('name', 'pvbase', k), '')
+            displayList.append(QStruct(v, pv, "temperature"))
+    # logger.debug("displayList", displayList)
```
ZLLentz, line 219: nitpick: hanging debug comment, imo you can either uncomment it and leave it in or delete (or ignore the nitpick)

ZLLentz (commented):
This is clean- just one more documentation suggestion and a minor minor nitpick

hutch_python/epics_arch.py
```diff
@@ -28,7 +49,16 @@ def _create_parser():
...
     parser.add_argument('--dry-run', action='store_true', default=False,
                         help='Print to stdout what would be written in the '
-                        'archFIle.')
+                        'archFile.')
+
+    parser.add_argument('--level', '-l', required=False, type=str, default="INFO",
+                        help='Change the logging level, e.g. DEBUG to show the debug logging stream')
+
+    parser.add_argument('--cds-items', nargs=2, action='store', default=None,
+                        help="Pulls all data from CDS tab in the form of a dictionary. E.g.: xppx1003221 run21 X-10032. This option will not automatically update the archfile.")
```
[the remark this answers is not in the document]
ZLLentz: Yes, that's much more clear! Thank you!

ZLLentz (approved):
Re-running the failed job (looks like an intermittent cloud networking error)
Otherwise this looks great, marking as approved!

ZLLentz:
OK I am going to merge this!
Deployment-wise: I can put this in dev_conda, and I can also tag the repo so it makes it into the next pcds_conda. Do you want either or both of these things?

PR #379, opened 2024-04-14
ENH: add option to limit devices loaded via `load_level` conf.yaml key

tangkong:
## Description
Adds `load_level` as a conf.yaml key that allows hutches to choose how many ophyd devices get loaded.  Currently includes 3 options:

- ``UPSTREAM``: The hutch's devices, and devices upstream from the requested hutch.
If there are multiple paths to the requested hutch, all paths' devices are loaded.
- ``STANDARD``: Devices gathered via ``UPSTREAM``, plus devices that share the
"beamline" field in happi with the ``UPSTREAM`` devices.  (The current standard)
- ``ALL``: All devices in the happi database.  Use this option at your own risk.

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-5079

Summary: We found that some hutches were loading many irrelevant devices.  TMO was catching CXI devices, XPP was catching XRT devices, etc.  This was found to be a result of the beamline field matching procedure.  

At some point I'd love to add a "hutch only" option, but there's no simple way to implement that currently.  This framework should support it though.

## How Has This Been Tested?
Interactively with a tmopython profile, added a test case

This PR, the jira ticket

## Pre-merge checklist

### DIFF
--- a/hutch_python/happi.py
+++ b/hutch_python/happi.py
@@ -1,3 +1,4 @@
+import enum
 import inspect
 import logging
 
@@ -17,10 +18,18 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
+class DeviceLoadLevel(enum.IntEnum):
+    # HUTCH = 0  # TODO: Figure out how to gather only hutch devices
+    UPSTREAM = 1
+    STANDARD = 2
+    ALL = 3
+
+
 def get_happi_objs(
     db: str,
     light_ctrl: LightController,
     endstation: str,
+    load_level: DeviceLoadLevel = DeviceLoadLevel.STANDARD
 ) -> dict[str, ophyd.Device]:
     """
     Get the relevant items for ``endstation`` from the happi database ``db``.
@@ -55,6 +64,11 @@ def get_happi_objs(
         client = happi.Client(path=db)
     containers = list()
 
+    if load_level == DeviceLoadLevel.ALL:
+        results = client.search(active=True)
+        containers.extend(res.item for res in results)
+        return _load_devices(*containers)
+
     if light_ctrl is None or (endstation.upper() not in light_ctrl.beamlines):
         # lightpath was unavailable, search by beamline name
         reqs = dict(beamline=endstation.upper(), active=True)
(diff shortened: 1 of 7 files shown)


hutch_python/tests/happi_db.json
ZLLentz, line 1: ascii diagram I made to help me understand the test updates
```
   |    TST    |  STS  | ZZZ |
X0 - 1 - 2 - 3 - 4 - 5 -
Z0                    \- 6
```

hutch_python/tests/test_happi.py
```diff
@@ -23,7 +23,7 @@ def test_happi_objs():
     # Only select active objects
     lc = get_lightpath(db, 'tst')
     objs = get_happi_objs(db, lc, 'tst')
-    assert len(objs) == 2
+    assert len(objs) == 4
```
ZLLentz, line 26: It looks to me like we only have 3 devices on tst- where does the 4th come from? Maybe I misunderstand, I'll re-read the function text.
EDIT: I see this matches the "STANDARD" load below, which means it's consistent.
Does this mean that "STANDARD" doesn't take into account the branching device 5 at the end?

hutch_python/tests/test_happi.py
```diff
@@ -34,6 +34,25 @@ def test_happi_objs():
             lc = get_lightpath(tmp.name, 'tst')
 
 
+@pytest.mark.parametrize('load_level, num_devices', [
+    (DeviceLoadLevel.UPSTREAM, 3),
+    (DeviceLoadLevel.STANDARD, 4),
+    (DeviceLoadLevel.ALL, 5)
```
ZLLentz, line 40: These don't quite match my understanding of the levels.
Upstream - I agree it should be 3 since that's all the TST devices
Standard - shouldn't this be 5, including devices 4 and 5 on X0?
All - shouldn't this be 6, including the Z0 device?

hutch_python/happi.py
```diff
@@ -73,21 +87,22 @@ def get_happi_objs(
...
-        reqs = dict(beamline=line, active=True)
-        results = client.search(**reqs)
-        blc = [res.item for res in results
-               if res.item.name not in dev_names]
-        # Add the beamline containers to the complete list
-        if blc:
-            containers.extend(blc)
+    if load_level >= DeviceLoadLevel.STANDARD:
+        # also any device with the same beamline name
+        # since lightpath only grabs lightpath-active devices
```
ZLLentz, line 92: I missed this justification previously, I wonder if this would cause any issues for non-lightpath devices in the hutches if we move to UPSTREAM mode. Do we need to have any sort of special handling for these kinds of devices?

ZLLentz (commented):
Code look good
Maybe I'm a bit confused on the definitions of the various modes vs what the test suite shows + I've just noticed the original justification for the extra loads and wonder if we need to be more nuanced here

hutch_python/tests/happi_db.json
tangkong: Looking at only the active devices I have:
```
source_X0 -- tst_device_5(STS) -- tst_device_1(TST) -- tst_device_2(TST) -- X0
                               \_ tst_device_4(STS) -- tst_device_6(ZZZ) -- Z0
```

Mostly mimicking the TMO/CXI situation

tangkong: I should have re-numbered them based on their actual z position, but I was too lazy.  That's my bad

hutch_python/tests/test_happi.py
tangkong: I think the updated map I posted should clear this up, but the main difference is that `tst_device_3` is inactive, and isn't seen

hutch_python/happi.py
tangkong: It's a good point.  My inclination is that if hutches want those devices they should just pick `STANDARD`, since I don't know how to find / filter a level between `STANDARD` and `UPSTREAM`

Something like this might be nice, but inaccessible given our current metadata
`UPSTREAM`
`UPSTREAM + lightpath-inactive-but-on-beampath`
`STANDARD`

hutch_python/tests/happi_db.json
ZLLentz: ahhh I didn't even check the z

ZLLentz: This resolves all the test suite confusion

hutch_python/happi.py
ZLLentz: That's a good counterpoint- how would we even know which devices to pick? Needs more thought.

ZLLentz (approved):

PR #383, opened 2024-05-29
Jira ticket ECS-5104: Automatically close hutch-python sessions that have been idle for 48 hours or more.

janeliu-slac:
## Summary
This PR adds a timer that automatically closes a hutch-python session that has been idle for 48 hours or more. The timer is implemented as a class called IPython Session Timer that runs in a daemon thread. Using the 'post_run_cell' hook a timestamp is created after every user interaction in the interpreter. This timestamp is compared with the maximum amount of time a session can be idle. If 48 hours have passed, the thread will close the hutch-python session.

## Current Issues
I wasn't able to find a way for the ipython interpreter to exit cleanly. I was hoping to get the main thread (the ipython interpreter) to go through standard shutdown procedures and exit the program. The best solution I found is to use  `get_ipython().ask_exit()`. Other methods I tried caused problems (more details in section below).

After `get_ipython().ask_exit()` runs, the interpreter will display a second prompt. When the user types something at the prompt, ipython will run the code and then exit. This behavior doesn't seem ideal. This can be tested out by running hutch-python in my rhel7 folder at /cds/home/j/janeliu/git/hutch-python.

## Other methods to close ipython
I tried a couple of other methods to close the interpreter:

`os.kill(os.getpid(), signal.SIGINT)` - this raises a KeyboardInterrupt signal and is supposed to allow the main thread to follow standard shutdown procedures, but it generates an error message "KeyboardInterrupt escaped interact()" in hutch-python and the interpreter won't exit.

`os._exit(1)` - this command doesn't go through standard shutdown procedures. It does force ipython to exit, but my terminal freezes up afterwards. I think it's incorrectly terminating other processes.

This feature was discussed by maintainers of ipython, but it doesn't seem like it was ever implemented: https://github.com/ipython/ipython/issues/9944

## Pre-merge checklist

### DIFF
--- a/hutch_python/ipython_session_timer.py
+++ b/hutch_python/ipython_session_timer.py
@@ -6,6 +6,7 @@ maximum idle time can be updated. Currently, it is set to 48 hours.
 """
 
 import time
+import sys, select
 from IPython import get_ipython
 from threading import Thread
 
@@ -43,7 +44,7 @@ class IPythonSessionTimer:
 
     def __init__(self, ipython):
         self.curr_time = 0
-        self.max_idle_time = 172800
+        self.max_idle_time = 30  # 86400 is number of seconds in 24 hours
         self.last_active_time = 0
         self.idle_time = 0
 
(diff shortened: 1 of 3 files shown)


tangkong:
A couple of quick notes before diving into code (really just explaining the PR checklist):
* We usually advocate working on a named branch on your fork, rather than the master branch.  This isn't a showstopper but it'll help you keep organized
* There are some pre-commit checks that are failing.  These help keep style consistent across our repositories, as well as doing some basic static analysis.  You can install these with `pre-commit install` inside the top-level of a repository, and they'll run before you commit anything locally.  (pre-commit is installed in pcds_conda, but is easily pip-installable inside your local environment too)
* Another little thing is pre-release notes.  This helps us auto-generate detailed release notes with every tag.  run `./docs/pre-release-notes.sh <issue number> <description, typically branch name>` and fill out the generated file.

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,93 @@
...
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
+            self._timer(self.max_idle_time - self.idle_time)
+            self._get_time_passed()
+
+        # Close the user session
+        print("This hutch-python session has timed out. Please start a new session.")
+
+        # Close this ipython session
+        get_ipython().ask_exit()
```
tangkong, line 74: I'm also not sure if there's a better way.  Currently if the session times out, we can actually still request a close with `Ctrl+D`, though it doesn't matter what we choose.
image

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,93 @@
...
+def load_ipython_extension(ipython):
+    """
+    Initialize the `IPythonSessionTimer`.
+
+    This starts a timer that checks if the user session has been 
+    idle for 48 hours or longer. If so, close the user session.
+
+    Parameters
+    ----------
+    ip: ``ipython`` ``Shell``
```
tangkong, line 86: I know this is copied from `ipython_log.py`, but the parameter name is wrong and the type hint is... odd?  It should probably be the 
```
ipython : ``IPython.terminal.interactiveshell.TerminalInteractiveShell``
```
from above

tangkong (commented):
Overall I think this works as intended, at least in my interactive testing

Some tests would be good here, possibly monkeypatching the max_time to speed the timeout process up.

hutch_python/ipython_session_timer.py
ZLLentz: Taking inspiration from the linked github thread, we apparently need to follow this up with another exit call to kill the prompt that's waiting for input, otherwise as you say in the desc this waits for user input before exiting.
```
ip = get_ipython()
ip.ask_exit()
ip.pt_app.app.exit()
```

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,93 @@
...
+        time.sleep(sleep_time)
+
+    def _start_session(self):
+
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
+            self._timer(self.max_idle_time - self.idle_time)
+            self._get_time_passed()
+
+        # Close the user session
```
ZLLentz, line 70: I wonder if there's some way we can warn the user at regular intervals, e.g. "this session will time out in 2 hours if there is no further input"

hutch_python/ipython_session_timer.py
ZLLentz: The weird thing about IPython is that there are dozens of shell types

PR #384, opened 2024-06-20
Updating --cds-items option in epicsqarch-qs

ctslac:
## Description

I updated --cds-items to include a parser that helps the users find their experiments better. It handles a few different formats for the experiment name and if it can't find it right away it allows the user to input in the relevant information.
## Motivation and Context

It was requested by Vincent, the previous version required you to know the run number and proposal id. This way is slightly more interactive and allows you to 'search' for the experiment.
## How Has This Been Tested?

I didn't do any unit testing. I just tested it will all of the proposal id formats that I saw in the questionnaire and what was returned from getProposalsList from psdm_qs_cli.
## Pre-merge checklist

### DIFF
--- a/.github/workflows/standard.yml
+++ b/.github/workflows/standard.yml
@@ -20,7 +20,8 @@ jobs:
       # Extras to be installed only for conda-based testing:
       conda-testing-extras: "ipython=8.4.0"
       # Extras to be installed only for pip-based testing:
-      pip-testing-extras: "ipython==8.4.0"
+      # Set numpy<2 for now, remove when xraylib updates
+      pip-testing-extras: "ipython==8.4.0 numpy<2"
       # System packages to be installed only for conda-based testing:
       conda-system-packages: ""
       # System packages to be installed only for pip-based testing:
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -4,3 +4,5 @@ sphinx
 sphinx_rtd_theme
 sphinxcontrib-jquery
 pypandoc
+# Temporary pin, xraylib 4.1.3 (pcdscalc) incompatible
+numpy <2
--- a/docs/hutch_python.ipynb
+++ b/docs/hutch_python.ipynb
@@ -1,0 +1,496 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "be4d831f-7497-4a5d-b726-c70de60ce908",
+   "metadata": {},
+   "source": [
+    "# How to Use This Notebook\n",
+    "This notebook shows how to set up a jupyter notebook to load resources that were designed for use in the hutch-python ipython shell.\n",
+    "You should follow the instructions in the next cell to set up the server process, and then cherry-pick the following cells as needed based on what you want the user to be able to do in the notebook. Feel free to customize and experiment as needed.\n",
+    "\n",
+    "This was written assuming usage in xpp. You should be able to do exactly the same thing for other instruments, replacing references to xpp items with the relevent ones from another hutch."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7029ed26-b999-4eaa-8989-d5cb226a12e2",
+   "metadata": {},
+   "source": [
+    "# Setting up the jupyter server\n",
+    "First, you need to source xppenv (or xcsenv or etc.) to get the right env variables for the jupyter kernel.\n",
+    "This makes the jupyter kernel find the xpp beamline files and makes the DAQ work.\n",
+    "Note that this locks your terminal session to being compatible with exactly one particular version of the DAQ.\n",
+    "```\n",
+    "zlentz@psbuild-rhel7-02:~$ source /cds/group/pcds/pyps/apps/hutch-python/xpp/xppenv\n",
+    "Loading NFS python env pcds-5.8.0\n",
+    "```\n",
+    "\n",
+    "This also lets you register the python env for use in the notebook if you haven't already done this:\n",
+    "```\n",
+    "(pcds-5.8.0)zlentz@psbuild-rhel7-02:~$ python -m ipykernel install --user --name=pcds-5.8.0\n",
+    "Installed kernelspec pcds-5.8.0 in /cds/home/z/zlentz/.local/share/jupyter/kernels/pcds-5.8.0\n",
+    "```\n",
+    "\n",
+    "Then you can run the notebook from the same session:\n",
+    "```\n",
+    "(pcds-5.8.0)zlentz@psbuild-rhel7-02:~$ jupyter-notebook\n",
+    "```\n",
+    "\n",
+    "If you want to control the DAQ, then the jupyter-nootback process must be run on the same linux host that the DAQ is running on."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3a12ce95-bf6d-4742-a7c5-356aab40f743",
+   "metadata": {},
+   "source": [
+    "## Making sure it works\n",
+    "This cell should import without errors and there should be lots of \"xpp\" on the paths. Mine is slightly different since I'm running on psbuild instead of xpp-daq:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "9117cd79-d0e0-455a-8184-ef22af02b96f",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "/cds/group/pcds/pyps/apps/hutch-python/device_config/happi.cfg\n",
+      "/cds/group/pcds/pyps/apps/hutch-python/xpp:/cds/group/pcds/pyps/apps/hutch-python/xpp/dev/devpath:/cds/group/pcds/pyps/apps/hutch-python/xpp/../common/dev/devpath:/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/ami-current/build/ami/lib/x86_64-rhel7-opt\n",
+      ":/reg/g/pcds/dist/pds/xpp/ami-current/build/ami/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/ami-current/build/gsl/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/ami-current/build/qt/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/current/build/psalg/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/current/build/pds/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/current/build/pdsdata/lib/x86_64-rhel7-opt:/reg/g/pcds/dist/pds/xpp/current/build/offlinedb/lib/x86_64-rhel7-opt\n",
+      "/cds/group/pcds/pyps/apps/hutch-python/xpp/xpp/__init__.py\n",
+      "/reg/g/pcds/dist/pds/xpp/current/build/pdsapp/lib/x86_64-rhel7-opt/pydaq.abi3.so\n"
+     ]
+    }
+   ],
+   "source": [
+    "import os\n",
+    "import pydaq\n",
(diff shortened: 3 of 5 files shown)


ZLLentz:
Don't worry about the pip build failures, there's some disturbance in the python-ci-sphere right now due to some major package releases. I'll work through these in a follow-up PR.

I'll review the contents of your PR now.

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
 
-def pull_cds_data(exp, run):
+def pull_cds_items(exp):
+    logger.debug("in client")
+
     """
-    Gather all user objects from the CDS tab in the questionnaire.
-    Parse objects and separate them based on type.
-    Display them in the console via PrettyTable.
+    Gather all user obejcts from the CDS tab in the questionnaire.
```
tangkong, line 163: Did you edit these an intentionally mispell them?...

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+    create Pretty Table instance and if the values from the run data contain pcdssetup
+    then put them into a seperate dictionary as they are cds items
+    """
+
+    logger.debug('pull_cds_items:', exp)
+    try:
+        client = QuestionnaireClient()
+    except GSSError:
+        print('No kerberos token found; rerun after running kinit')
+        sys.exit()
```
ZLLentz, line 191: You should avoid calling `sys.exit` in functions other than the main entry point, else someone who tries to import and use this elsewhere might have their whole program close unexpectedly, when maybe they just want to know that this failed and do something else- perhaps in a gui for example.

It should be sufficient to return here, or maybe re-raise the exception?

```suggestion
        return
```

```suggestion
        raise
```

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+
+    formatted_run_id = ''
+    run_num = ''
+
+    # handle formatting of proposal id, want: e.g. X-10032
+    # Case -  Expected Format: e.g. xppx1003221
+    if re.match('^.{4}[0-9]{7}$', exp):
```
ZLLentz, line 198: Reviewer's note: here's what this regex means according to regex101

image

ZLLentz: I'm going to try it on all the active experiment names from the elog as part of the review:
https://regex101.com/r/JuWNa1/1

It looks like this works for all the new experiment names but fails a bit on the commissioning experiments. It's probably sufficient but you can consider allowing variable length numeric sections.

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+        logger.debug("experiment name format")
+        run_num = 'run'+exp[-2:]
+        logger.debug('run num', run_num)
+        run_id = str(exp[3:-2])
```
ZLLentz, line 202: This works because you only need the bits from the start/end of a fixed-length string and you shouldn't change it, but you should be aware of capturing groups and how you can use these to handle arbitrarily complex regex matching/extraction. See https://docs.python.org/3/library/re.html

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+        logger.debug('run_id', run_id)
+        formatted_run_id = run_id.capitalize()
+        formatted_run_id = formatted_run_id[:1] + '-' + formatted_run_id[1:]
+
+    elif re.match('[A-Z]{1}-[0-9]{5}$', exp) or re.match('^[A-Z]{2}[0-9]{2}$', exp):
```
ZLLentz, line 207: Reviewer note: one letter, a dash, and 5 numbers, or two letters and two numbers

ZLLentz: This does match both the new and the old proposal-id-only format :+1:

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+        runDetails_Dict = client.getProposalsListForRun(run_num)
+        for key, vals in runDetails_Dict.items():
+            logger.debug(formatted_run_id, key)
+            if str(key) == str(formatted_run_id):
+                logger.debug('matched key', key)
+                matchedKey = key
+
+    except Exception as e:
+        print("An invalid https request, please check the run number, proposal id and experiment number:", e)
+        sys.exit()
```
ZLLentz, line 233: Consider again logging exceptions and returning or raising instead of exiting the entire program
```suggestion
        logger.error("An invalid https request, please check the run number, proposal id and experiment number:", e)
        return
```

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
+
+    try:
+        runDetails_Dict = client.getProposalDetailsForRun(run_num, matchedKey)
+        # questionnaireFlag = True
+    except (Exception, UnboundLocalError) as e:
+        print('Could not find experiment, please check to make sure information is correct.', e)
+        sys.exit()
```
ZLLentz, line 240:
```suggestion
        logger.error('Could not find experiment, please check to make sure information is correct.', e)
        return
```

hutch_python/epics_arch.py
```diff
@@ -219,6 +280,7 @@ def pull_cds_data(exp, run):
         elif re.match('pcdssetup-temp.*-name', k):
             pv = cds_dict.get(re.sub('name', 'pvbase', k), '')
             displayList.append(QStruct(v, pv, "temperature"))
+    # logger.debug("displayList", displayList)
```
ZLLentz, line 283: nitpick: hanging comment, it's usually ok to leave debugs uncommented (unless this is huge?)

ZLLentz (commented):
I like this a lot!

I have a ton of comments about maintenance/code health (thinking about the lifetime of the code beyond this PR) and not a lot about your intended changes, which look good.

hutch_python/epics_arch.py
ZLLentz: I'm assuming this is a git mishap from a previous PR where we fixed typos as in-line suggestions in github but they didn't propagate to his fork

hutch_python/epics_arch.py
ZLLentz: No need to adjust this here! If it works, it works! (If it doesn't work, there's my suggestion though!)

hutch_python/epics_arch.py
```diff
@@ -7,6 +7,7 @@
 import sys
 from dataclasses import dataclass
 
+from kerberos import GSSError
```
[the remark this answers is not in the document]
ZLLentz: Think of kerberos as a centralized authentication service. Every normal user on each of our servers on the network should be able to get kerberos tokens.

That being said, hutch *opr users don't normally get kerberos tokens... Maybe we need to do something special for these again? In the past these have accessed the questionnaire via hard-coded read-only username/password pairs.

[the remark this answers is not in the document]
ZLLentz: In the past we passed in a username and password into the init for `QuestionnaireClient` that came from a file in the opr home area: 
- https://github.com/pcdshub/psdm_qs_cli/blob/9d6bed253aa3a9f2c49f2fe43d3402b8cd9a6042/psdm_qs_cli/QuestionnaireClient.py#L49
- https://github.com/pcdshub/hutch-python/blob/a2b1b55a2fb83da9ebdf82d4a9ec48f54ac6cc87/hutch_python/qs_load.py#L85-L94

hutch_python/epics_arch.py
ZLLentz: If it's huge and you feel like you may need to bring it back frequently it's ok to keep it

hutch_python/epics_arch.py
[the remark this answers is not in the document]
ZLLentz: iirc all the opr home areas have this file
check out your favorite opr's home and find `.web.cfg`

```
$ ll /cds/home/opr/*/.web.cfg
-rw-rw-r-- 1 cxiopr cxiopr 63 Aug 27  2020 /cds/home/opr/cxiopr/.web.cfg
-rw-rw-r-- 1 mecopr mecopr 63 Nov  6  2023 /cds/home/opr/mecopr/.web.cfg
-rw-r--r-- 1 mfxopr mfxopr 63 Oct 16  2020 /cds/home/opr/mfxopr/.web.cfg
-rw-rw-r-- 1 rixopr xs     63 Apr 20  2021 /cds/home/opr/rixopr/.web.cfg
-rw-rw-r-- 1 tmoopr ps-tmo 63 Apr 15  2021 /cds/home/opr/tmoopr/.web.cfg
-rw-r--r-- 1 xcsopr xcsopr 63 Aug 27  2020 /cds/home/opr/xcsopr/.web.cfg
-rw-r--r-- 1 xppopr xppopr 63 Aug 27  2020 /cds/home/opr/xppopr/.web.cfg
```

conda-recipe/meta.yaml
```diff
@@ -43,6 +43,7 @@ requirements:
   - pyyaml >=5.4.0
   - requests >=2.26.0
   - simplejson >=3.17.0
+  - pykerberos
```
ZLLentz, line 46: small nitpick: the packages in meta.yaml and requirements.txt are in alphabetical order aside from this addition

hutch_python/epics_arch.py
ZLLentz: I think this is a good change, it makes it more robust against variable-length numeric sections now and in the future

conda-recipe/meta.yaml
ZLLentz, line 1: One small annoying thing: the pip and docs builds are failing to install `pykerberos` because they need an additional system package.
Can you update the line here to include `"libkrb5-dev"` instead of `""`?
https://github.com/pcdshub/hutch-python/blob/00e39a3c29feb774f33bcf2cb0edea3ff25d4328/.github/workflows/standard.yml#L28

ZLLentz:
Hey the tests are running again! Great!
I'm going to go through this one more time as a reviewer

hutch_python/epics_arch.py
```diff
@@ -156,33 +156,93 @@ def create_arch_file(experiment, level=None, hutch=None, path=None, dry_run=Fals
...
         The experiment's name e.g. xppx1003221
     run: ''str''
         The run number e.g. run21
-
+    ----------
     Outputs
     -------
         PrettyTable visualization of cds objects
+    -------
+
```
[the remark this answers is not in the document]
ZLLentz: This extra line of hyphens still should be removed, it will likely break the auto-docs formatting

hutch_python/epics_arch.py
```diff
@@ -172,17 +174,94 @@ def pull_cds_data(exp, run):
...
+        except NoOptionError:
+            logger.error("Could not find valid username and password in qs.cfg or web.cfg, attempting to autheticate with kerberos instead.")
+    else:
+        logger.debug("Cfgs could not be found. ")
+    if client is None:
+        try:
+            client = QuestionnaireClient(use_kerberos=True)
+        except Exception as exc:
+            logger.error("Not able to authenticate with kerberos, please make sure you have an active token.")
+            raise exc
```
ZLLentz, line 210: I like this implementation of the pw/kerb auth :+1:

requirements.txt
```diff
@@ -18,6 +18,7 @@ pyqt5
 pyyaml>=5.4.0
 requests>=2.26.0
 simplejson>=3.17.0
+pykerberos
```
ZLLentz, line 21: this file should also be alphabetical for maintainability

ZLLentz:
By and large this is very good, just the two remaining things:
- The extra row of `------------` in the structured docstring that will break the html generation
- The package order in `requirements.txt`

ZLLentz (approved):
I'm happy with this!
If you're happy and finished, feel free to mark this as "ready for review" (currently marked as a draft) and click merge if you want (or I'll click if for you)
You should also click the "Update branch" button

ZLLentz:
Yes, I think you've done your full due diligence with your regex filters. It could miss an oddly named one here or there but there's only so much we can do with finite time. I suspect it wouldn't make sense to change them again unless we get another new naming convention.

PR #385, opened 2024-06-25
ECS-5104 Hutch-python feature: auto-close after timeout

janeliu-slac:
## Description
PR for the IPython session timer

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-5104

## How Has This Been Tested?
Haven't been able to test changes to load_config.py and rix/conf.yml. How are changes to <hutch>/conf.yml file tested?

Documentation is in the file comments.

## Pre-merge checklist
- [x ] Code contains descriptive docstrings, including context and API


tangkong:
There are some pre-commit fixes to address here.  If you install pre-commit locally you can catch these before pushing.

Some weirdness in CI here.  Looks like the passing tests grab an older version of pcdscalc, which doesn't depend on xraylib?

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,99 @@
...
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
+            self._timer(self.max_idle_time - self.idle_time)
+            self._get_time_passed()
+
+        # Close the user session
+        print("This hutch-python session has timed out and automatically closed. Any code that is running will continue to run until it is completed.")
+
+        # Close this ipython session
+        get_ipython().ask_exit()
```
ZLLentz, line 81: We apparently need to follow this up with another exit call to kill the prompt that's waiting for input, otherwise as you say in the desc this waits for user input before exiting.
```
ip = get_ipython()
ip.ask_exit()
ip.pt_app.app.exit()
```

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,99 @@
...
+from threading import Thread
+
+from IPython import get_ipython
+
+max_idle_time = 172800
+
+
+def configure_timeout(duration):
+    if isinstance(duration, int) and duration > 0:
+        max_idle_time = duration
```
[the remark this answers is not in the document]
janeliu-slac: I had this in my vs code and forgot to update in the rhel7 version. Fixed.

hutch_python/ipython_session_timer.py
janeliu-slac: `ip.pt_app.app.exit()` seems to generate an error message when the user code is running. I wasn't able to find another clean way to kill the prompt and exit ipython.

From what I observed the 'post_run_cell' hook triggers after the user enters something at the prompt. It doesn't trigger when the user's code is running. So, the self.last_active_time variable is set to the time that I entered something at the prompt and not when my user code is running.

In the screenshot 'post_run_cell' gets a timestamp of when I first pressed enter at the prompt (for example, at time 00:00:00). Then the timer starts to count down. After 20 seconds the timer script checks self.last_active_time, which still shows 00:00:00, so it ends the session. So, the session timeout message shows up 20 seconds after I entered code at the prompt and not 20 seconds after my user code finishes running.

screenshot_of_exit_call

ZLLentz: Ahh I think this error is because the `ip.pt_app.app` object is only "running" while waiting for user input

ZLLentz: That is to say, my suggestion here is only valid for cancelling the prompt input itself (automatically) and is only valid when there is a prompt.

janeliu-slac: Yep, it worked well for me when there was no user code running.

janeliu-slac: I think this is why I thought about using zeromq or websockets to check the status of the ipython kernel and figure out if it's busy or idle. If the kernel is idle, then we'd be able to wait for the user code to finish running and ip.pt_app.app would get ipython to exit cleanly.

tangkong: Maybe I'm also misunderstanding things, but it seems to me that the `post_run_cell` runs after the cell is finished executing 
```ipython
In [4]: for i in range(5):
   ...:     time.sleep(1)
   ...:     print(i)
   ...:
pre-run
0
1
2
3
4
post-run
```

So we could switch off the timer at `pre_run_cell`, then record the last active time and start counting down on the `post_run_cell`?

janeliu-slac: From my observation `post_run_cell` is triggering after the first line `for i in range(5):`. When post_run_cell triggers it calls the `_set_active_time()` function, which sets `self.last_active_time` and the 20 second timer begins. When 20 seconds have passed the script checks how long the user has been idle (user has not entered anything at the prompt). If  self.last_active_time has not changed in the last 20 seconds the script prints out the "This hutch-python session has timed out..." message.

Adding to the confusion is that typing `time.sleep(1)` and `print(i)` doesn't seem to trigger `post_cell_run`. So the "This hutch-python session has timed out..." message is printed to the console just a few seconds after this loop runs.

I think my implementation might be a little convoluted / hard to read.

tangkong: So the snippet I posted was generated with this small modification to your code:
```python
        # _set_last_active_time() function will trigger every time user runs a cell
        ipython.events.register('pre_run_cell', self._set_last_active_time)
        ipython.events.register('pre_run_cell', lambda: print('pre-run'))
        ipython.events.register('post_run_cell', lambda: print('post-run'))
```

Which seems to conflict with your observation.  As currently written, this PR only registers an event to the `pre_run_cell` event right?

janeliu-slac: Right, this PR only registers an event to the  'pre_run_cell' hook. I was testing with 'post_run_cell' to see how it behaves. 

I think there might be an issue with the logic in my code.  Going to track down the bug.

janeliu-slac: I figured out the error in the script and rewrote the code so it's simpler. `ip.pt_app.app.exit()` should be working now. The test code has a 30 second timer and is in my development folder /cds/home/j/janeliu/git/hutch-python/hutch_python.

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _set_user_inactive(self, result):
+        self.user_active = False
+
+    def _start_session(self):
+        # poll for user activity
+        while (1):
+            if not self.user_active:
+                time.sleep(self.max_idle_time)
+                if not self.user_active:
```
tangkong, line 64: This logic works for long running cells, but doesn't "refresh" the time when a cell is quick.  Quick cells flip the active flag back and forth to false immediately, so the idle sleep timer from the previous cell can break this loop.

You can test this by running some long cell, then repeatedly doing some simple assignment.  

I think we actually need to keep track of the end time of the last completed cell as well, and timeout based on that as well.

docs/source/upcoming_release_notes/385-Hutch-python_feature:_auto-close_after_timeout.rst
```diff
@@ -0,0 +1,22 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
```
tangkong, line 22:
```suggestion
- janeliu-slac
```

docs/source/upcoming_release_notes/385-Hutch-python_feature:_auto-close_after_timeout.rst
```diff
@@ -0,0 +1,22 @@
+385 Hutch-python feature: auto-close after timeout
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
```
tangkong, line 10: We need to fill out these sections to help us set up the release notes later down the line

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,94 @@
+"""
+This module modifies an ``ipython`` shell to automatically close if it has been
+idle for 48 hours. Prior to starting hutch-python the user is alerted that the
```
tangkong, line 3: Technically now that you've let us configure this, this module docstring is out of date

tangkong:
I think we also need some unit tests here.  This is the kind of feature that can be super disruptive if we ever get some kind of regression

hutch_python/ipython_session_timer.py
janeliu-slac: I couldn't get the implementation using `pre_run_cell` and `post_run_cell` to cancel the threading.Timer() function. There is a stackoverflow post about using a global t = Timer() function but doing this with self.T = Timer() didn't work (https://stackoverflow.com/questions/56566801/python-timer-start-and-reset). I went back to the first iteration of ipython_session_timer.py and tried to create a combination that works. I think the latest commit should be working (hope nothing breaks!)  Will try to add some unit tests and fix other issues.

hutch_python/ipython_session_timer.py
```diff
@@ -36,41 +36,62 @@ class IPythonSessionTimer:
...
         self.user_active = False
 
-        # _set_last_active_time() function will trigger every time user runs a cell
         ipython.events.register('pre_run_cell', self._set_user_active)
         ipython.events.register('post_run_cell', self._set_user_inactive)
 
-    def _set_user_active(self, result):
+    def _set_user_active(self):
         self.user_active = True
+        self.last_active_time = time.time()
```
tangkong, line 66: funny python-y thing.  We should use `time.monotonic()` here, `time.time()` is the system time I believe, and can change.

hutch_python/ipython_session_timer.py
```diff
@@ -36,41 +36,62 @@ class IPythonSessionTimer:
...
+    idle_time: float
+        The amount of time the user session has been idle.
     '''
 
     def __init__(self, ipython):
         global max_idle_time
-        self.max_idle_time = max_idle_time
+        self.curr_time = 0
+        self.max_idle_time = 30  # max_idle_time
+        self.last_active_time = 0
```
tangkong, line 57: A bit of debugging settings leftover here I think. 
```python
self.max_idle_time = max_idle_time
self.last_active_time = time.monotonic()
```

hutch_python/ipython_session_timer.py
```diff
@@ -36,41 +36,62 @@ class IPythonSessionTimer:
...
 
-        # _set_last_active_time() function will trigger every time user runs a cell
         ipython.events.register('pre_run_cell', self._set_user_active)
         ipython.events.register('post_run_cell', self._set_user_inactive)
 
-    def _set_user_active(self, result):
+    def _set_user_active(self):
         self.user_active = True
+        self.last_active_time = time.time()
+        print("pre_cell_run")
```
tangkong, line 67: also some leftover prints

hutch_python/ipython_session_timer.py
```diff
@@ -36,41 +36,62 @@ class IPythonSessionTimer:
...
-
-        # Close the user session
-        print("This hutch-python session has timed out and automatically closed. Please start a new session")
-
-        # Close this ipython session
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
+
+            while (self.user_active):
+                time.sleep(6)
```
tangkong, line 86: Is there special significance to this number?  It's fine how it is, it just seems strangely specific

tangkong (commented):
I think the logic of this is looking a lot better 👍 Stray prints and debug settings aside

hutch_python/ipython_session_timer.py
janeliu-slac: It was for testing. It's supposed to check if the user is active once a minute but the test timeout duration was 20 seconds so I changed it to 6. It's been changed back to 60 seconds.

hutch_python/load_conf.py
```diff
@@ -202,7 +203,8 @@ def load_conf(conf, hutch_dir=None, args=None):
             db = None
     except KeyError:
         db = None
-        logger.info('Missing db from conf. Will skip loading from shared database.')
+        logger.info(
+            'Missing db from conf. Will skip loading from shared database.')
```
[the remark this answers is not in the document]
janeliu-slac: Just made this change and pushed to the repo.

docs/source/upcoming_release_notes/385-Hutch-python_feature:_auto-close_after_timeout.rst
janeliu-slac: Think I updated this correctly.

janeliu-slac:
> I think we also need some unit tests here. This is the kind of feature that can be super disruptive if we ever get some kind of regression

Writing some unit tests and reading up on PyTest.

docs/source/upcoming_release_notes/385-ECS-5104_Hutch-python_feature:_auto-close_after_timeout.rst
```diff
@@ -0,0 +1,22 @@
+385 ECS-5104 Hutch-python feature: auto-close after timeout
+#################
+
+API Changes
+-----------
+- Created an IPython extension that launches a daemon thread that runs an instance of IPythonSessionTimer class object.
```
tangkong, line 6: We usually use this section to denote breaking API changes, which I wouldn't classify this as.  Nothing you've done here breaks hutch-python's interactions with other packages / interfaces, since we use a default value if nothing in the user's hutch-python configuration changes

janeliu-slac: Ah ok, I removed this line.

janeliu-slac: I just realized how many minor commits I have in my features branch (mostly fixing errors). Are those commits squashed when merging the pull request?

tangkong: We can squash before merging.  I usually don't mind having all the commits but if there's a lot of back and forth we can

janeliu-slac:
I put together some unit tests for ipython_session_timer.py. There is probably a better way to test IPythonSessionTimer._start_session() but this was the only version I was able to get working.

hutch_python/tests/test_ipython_session_timer.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+# Test case 2: the user is inactive and their idle time is equivalent to or exceeds the maximum allowable idle time.
+@unittest.mock.patch('hutch_python.ipython_session_timer.time.sleep', lambda seconds: None)
+def test_start_session_case2(session_timer, capsys):
+    session_timer.idle_time = session_timer.max_idle_time
+
+    if (session_timer.idle_time >= session_timer.max_idle_time) and not session_timer.user_active:
+        captured = capsys.readouterr()
+        assert captured == "This hutch-python session has timed out. Please start a new session."
```
tangkong, line 58: A note for future tests, straight string equality checks like this can be a bit brittle
Less breakable, checking for a substring instead of the exact output:
```python
assert "timed out" in captured
```

Even more robust is mocking session_timer.ip.ask_exit and confirming it was called

hutch_python/tests/test_ipython_session_timer.py
```diff
@@ -0,0 +1,58 @@
...
+        session_timer.last_active_time = session_timer.curr_time - 200.0
+        assert session_timer.idle_time == 200.0
+
+
+# Test case 2: the user is inactive and their idle time is equivalent to or exceeds the maximum allowable idle time.
+@unittest.mock.patch('hutch_python.ipython_session_timer.time.sleep', lambda seconds: None)
+def test_start_session_case2(session_timer, capsys):
+    session_timer.idle_time = session_timer.max_idle_time
+
+    if (session_timer.idle_time >= session_timer.max_idle_time) and not session_timer.user_active:
```
tangkong, line 56: I just realized that this will always evaluate `False`, since you've set `session_timer.user_active=True` in the fixture

tangkong (commented):
I think we could refine the tests, but perhaps we leave that for a follow up.  We've sat on this long enough.  One test I'd like to add later on is one that actually tests the timer functionality.  As in: we set the timeout time to ~1s, and watch for the exit signal to be sent.  But that can be a later problem.  

The last two tests also don't actually assert anything, the if statements always resolve to False

I think interactively this works as advertised, and the default being at 48 hours is good enough to be minimally intrusive.

hutch_python/tests/test_ipython_session_timer.py
tangkong: In fact, since the session timer never actually gets kicked off, the loop that evaluates this never runs.  So even if we do enter this loop nothing gets captured in `capsys.readouterr`

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,108 @@
...
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
+
+            # Check if user is active once every minute
+            while (self.user_active):
+                time.sleep(60)
+                self.idle_time = 0
+
+            time.sleep(self.max_idle_time - self.idle_time)
+            self._set_idle_time()
```
ZLLentz, line 84: Did using `threading.Timer` not work out? I thought that was the cleanest approach.

I'm trying to step through the logic in my head to see if it makes sense to me, since I'm finding it unintuitive.

- At the start, we initialize `self.max_idle_time` to some large configurable, `self.idle_time` to zero, and `self.user_active` to `False`
- Outer while loop evaluates to `True` at first
- Inner while loop evaluates to `False` at first
- We sleep for a long, long time (the configured `self.max_idle_time`)
- While waiting, some number of `_set_user_active` calls are made (which are followed by `_set_user_inactive` calls), which flip the `user_active` boolean to `True` and then to `False` respectively
- Either way, during each of these calls we're updating `self.last_active_time` to the current time
- After the long sleep, we update the idle time to the number of seconds since the last `_set_user_active` or `_set_user_inactive` call
- Now, we have the first chance to exit the loop, which will happen if it's been long enough since `_set_user_active` or `_set_user_inactive`, or if neither has ever been called.
- Suppose we stay in the loop, then we either have `self.user_active` set to `True`, in which case we do a small loop of re-checking for this every minute until the user command ends, successfully preventing us from exiting, or we drop back into the big sleep (which may be less big this time depending on the situation).

Despite confusing me, I think this generally works. There's a small chance of a race condition if a user starts a long command right before the start of the big sleep on line 83, but otherwise this works out reasonably well.

I think I probably would have opted for creating and interrupting timer objects, because it wouldn't have required maintaining state across while loop iterations, it would just have required me to kill a timer at `_set_user_active` and remake it again at `_set_user_inactive`.

hutch_python/tests/test_ipython_session_timer.py
ZLLentz, line 1: None of the tests in this file start the session timer

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,108 @@
...
+
+    last_active_time: float
+        The time of the last user activity in this session.
+
+    idle_time: float
+        The amount of time the user session has been idle.
+    '''
+
+    def __init__(self, ipython):
+        global max_idle_time
```
ZLLentz, line 51: nitpick 1: no need to declare a global if you don't need to modify it. The scope here can read the variable with no issues without the `global` keyword.
```suggestion
```

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,108 @@
...
+    idle for 48 hours or longer. If so, close the user session.
+
+    Parameters
+    ----------
+    ipython: IPython.terminal.interactiveshell.TerminalInteractiveShell
+        The active ``ipython`` ``Shell``, the one returned by
+        ``IPython.get_ipython()``.
+    """
+    UserSessionTimer = IPythonSessionTimer(ipython)
+    t1 = Thread(target=UserSessionTimer._start_session, daemon=True)
```
ZLLentz, line 107: nitpick 2: python naming conventions in pep8 (https://peps.python.org/pep-0008/) specifies `CamelCase` for classes, but `UserSessionTimer` is a variable, so it should follow the "function and variable names" convention and be user_session_timer
```suggestion
    user_session_timer= IPythonSessionTimer(ipython)
    t1 = Thread(target=user_session_timer._start_session, daemon=True)
```

ZLLentz (commented):
Consolidating small nitpick suggestions into one group here

hutch_python/ipython_session_timer.py
tangkong: There was some comments about the timer not working but it lacks some details https://github.com/pcdshub/hutch-python/pull/385#discussion_r1655729172

janeliu-slac: Yep, I did have an implementation where threading.Timer() is launched by `post_run_cell` and canceled by `pre_run_cell`. I couldn't get this implementation to cancel and relaunch Timer() (it looks like threading.Timer() is launched as a thread so once it's canceled a new one has to be spawned). 

I found a stackoverflow solution using a global t = Timer() function to launch and cancel Timer(), but because it gets called inside the IPythonSessionTimer class I tried using self.T = threading.Timer() to substitute for global t. Unfortunately, I couldn't get this to work, so I went back to the first iteration of ipython_session_timer.py and created a combination that works. Unfortunately, the implementation is kind of unintuitive.

ZLLentz: I see, it's really strange that creating a new timer with self.T wouldn't work.

hutch_python/ipython_session_timer.py
```diff
@@ -0,0 +1,108 @@
...
+        self.user_active = False
+        self.last_active_time = time.monotonic()
+
+    def _set_idle_time(self):
+        self.curr_time = time.monotonic()
+        self.idle_time = self.curr_time - self.last_active_time
+
+    def _start_session(self):
+        # Check if idle_time has exceeded max_idle_time
+        while (self.idle_time < self.max_idle_time):
```
ZLLentz, line 76: At the risk of complicating things further, here's my minimal suggestion for fending off any potential remaining race conditions with ongoing long commands:
```suggestion
        while (self.idle_time < self.max_idle_time) or self.user_active:
```

hutch_python/ipython_session_timer.py
janeliu-slac: This was my code for the implementation:
```
class IPythonSessionTimer:
    def __init__(self, ipython):
        global max_idle_time
        self.T = Timer(20.0, self._end_session)

        ipython.events.register('pre_run_cell', self.T.cancel)
        ipython.events.register('post_run_cell', self._start_timer)

    def _end_session(self):
        print("This hutch-python session has timed out and automatically closed. Please start a new session.")

        # Close this ipython session
        ip = get_ipython()
        ip.ask_exit()
        ip.pt_app.app.exit()

    def _new_timer(self):
        global max_idle_time
        print("Timer has started.")
        self.T = Timer(max_idle_time, self._end_session)

    def _start_timer(self):
        self.T.cancel()
        self._new_timer()
        self.T.start()

    def _start_session(self):
        self._new_timer()
```

ZLLentz: That's a real shame because the above snippet is clean and way easier to understand

ZLLentz: The only thing I see is that the original self.T is never started in the above snippet, only the subsequent timers are started

janeliu-slac:
```
def _start_session(self):
        self.T.start()
```
Tried this and no luck :\

ZLLentz: Yes, but `self._start_session()` is never called in the snippit.

janeliu-slac: Ah, I forgot to include the def load_ipython_extension(ipython) function in the code above. Updated.

janeliu-slac:
I added some code to `class FakeIPython` in test_ipython_log.py to avoid error messages for `ipython.ask_exit()` and `ipython.pt_app.app.exit()`. Just wanted to check if this looks right before I commit it to test_ipython_log.py.

```
class FakeIPython:
    ...

    def ask_exit():
        pass

    class pt_app:
        
        def __init__(self):
            pass

        class app:
            
            def __init__(self):
                pass

            def exit():
                pass
```

tangkong:
You could probably use `[unittest.mock.MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock)` to individually mock the items that you care about.  This object has a bunch of helpful methods for tracking calls

```python
from types import SimpleNamespace
from unittest.mock import MagicMock

class FakePtApp:
    app = SimpleNamespace(exit=MagicMock())

class FakeIPython:
    """A fake replacement of IPython's ``TerminalInteractiveShell``."""
    user_ns: dict[str, Any]
    events: FakeIPythonEvents

    def __init__(self):
        self.user_ns = dict(In=[""])
        self.events = FakeIPythonEvents()
        self.pt_app = FakePtApp()
        self.ask_exit = MagicMock()
```
later in the tests: 
```python
@unittest.mock.patch('time.sleep', lambda seconds: None)
def test_start_session_case1(session_timer, fake_ipython, capsys):
    session_timer.user_active = False

    session_timer._start_session()

    captured = capsys.readouterr()
    assert "timed out" in captured.out
    assert session_timer.ip.ask_exit.called
    assert session_timer.ip.pt_app.app.exit.called
```

Aside from that, the code snippet you shared doesn't look entirely right / complete.  You're creating the `pt_app` class but never instantiating it.  I think it technically works but it looks weird to me.  You should be able to run the test suite locally to verify things work.

janeliu-slac:
Thanks for the MagicMock suggestion. Going to read up on it and learn what it's doing. I've added this and the unit tests all pass. Does the hutch-python session timer feature look ready to merge?

hutch_python/tests/test_ipython_session_timer.py
```diff
@@ -0,0 +1,52 @@
...
+    assert session_timer.user_active
+
+
+def test_set_user_inactive(session_timer):
+    session_timer.user_active = False
+    assert not session_timer.user_active
+
+
+def test_set_idle_time(session_timer):
+    assert session_timer.idle_time == 10.0
```
ZLLentz, line 37: I'm not entirely sure what these three tests are for, it seems to me like they're checking if the fixture is set up as expected.

hutch_python/tests/test_ipython_session_timer.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+# Skipping tests for _start_session() where self.user_active==True because this enters an
+# infinite while loop.
```
ZLLentz, line 41: It's probably worthwhile finding a way to test and make sure that a session with `self.user_active` isn't exited. This might look something like running `_start_session` in a `threading.Thread` as done in `load_ipython_extension`. The infinite while loop only applies when this runs in the foreground thread. We could then set the max idle time to e.g. 1s and verify that ask_exit and exit never get called after 5s or so. Then, in the same test, we could flip `self.user_active` back to `False` and see that `ask_exit`, etc. get called soon after.

hutch_python/tests/test_ipython_session_timer.py
```diff
@@ -0,0 +1,52 @@
...
+@unittest.mock.patch('time.sleep', lambda seconds: None)
+def test_start_session(session_timer, fake_ipython, capsys):
+    session_timer.user_active = False
+
+    session_timer._start_session()
+
+    captured = capsys.readouterr()
+    assert "timed out" in captured.out
+    assert session_timer.ip.ask_exit.called
+    assert session_timer.ip.pt_app.app.exit.called
```
ZLLentz, line 52: I like this test, it gets to the core of what the feature is supposed to do and tests it about as directly as we can. Out of curiosity, how long does this take to run? Something like 5-10s is probably OK, much higher and we start to slow down the full test suite.

tangkong (dismissed):
I think this looks good to me.  Some hard fought lines of code but I like where it ended up! 🎉

ZLLentz (commented):
Three questions/thoughts about the unit tests
Overall this is great and we can probably merge soon

ZLLentz:
fwiw I think we could definitely merge as-is, it does the thing correctly and is well-documented. I'm splitting hairs on the unit tests at this point.

hutch_python/tests/test_ipython_session_timer.py
janeliu-slac: Oh, forgot about these. It's from when I didn't really understand how Pytest works. Fixed.

hutch_python/tests/test_ipython_session_timer.py
janeliu-slac: I lowered the session_timer.max_idle_time() to 5 seconds.

hutch_python/tests/test_ipython_session_timer.py
janeliu-slac: This might take some time to set up. Could we merge without this one?

ZLLentz: Yes, I think it's healthy to merge and consider this later (if ever)

ZLLentz (approved):

janeliu-slac:
cool, it's done! I'll merge tomorrow morning.

PR #386, opened 2024-06-27
CI: temporarily drop numpy to under 2 in pypi builds

ZLLentz:
## Description

Force the pypi builds to run with numpy<2

## Motivation and Context

Some of these builds are failing to no fault of their source PRs.

## How Has This Been Tested?

It worked on pcdsdevices

Here only

## Pre-merge checklist

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -4,5 +4,5 @@ sphinx
 sphinx_rtd_theme
 sphinxcontrib-jquery
 pypandoc
-# Temporary pin, can remove once it installs without this
+# Temporary pin, xraylib 4.1.3 (pcdscalc) incompatible
 numpy <2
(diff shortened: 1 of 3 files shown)


ZLLentz:
There isn't currently a spot in the .github config to set package overrides for the docs build, so I have put this into the docs requirements instead.

tangkong:
The most recent np2 related failure I've seen is a `pcdscalc` -> `xraylib` incompatibility.  We're going to need xraylib to update before this will ever work, and that package is unfortunately out of our purview.

docs-requirements.txt
```diff
@@ -4,3 +4,5 @@ sphinx
 sphinx_rtd_theme
 sphinxcontrib-jquery
 pypandoc
+# Temporary pin, can remove once it installs without this
```
tangkong, line 7:
```suggestion
# Temporary pin, xraylib 4.1.3 (pcdscalc) incompatible
```
I think future me will need more verbosity here

tangkong (commented):
Maybe we just pin this upstream in the ci helpers.  But recording the incompatibility here is probably also appropriate.  We can pin in both places

tangkong (approved):

PR #389, opened 2024-10-11
FIX: more error handling for ElogPoster

ZLLentz:
## Description

This comes from a hotfix in rix3 that I added onto here.

There is a niche issue where "something" gets loaded as the `elog` but it isn't a valid elog object. Then, this script picks up the non-elog item and passes it to the `ElogPoster`, which starts causing an issue (text spam? I don't quite remember) when you try to run a scan later.

This PR rearranges the error reporting a bit and includes a check that the expected objects haven't been replaced.

## Motivation and Context

- Working through hotfixes as part of https://jira.slac.stanford.edu/browse/ECS-6333
- Can cause an annoying error spam if the user makes an `elog` object, instead just fail the `ElogPoster` in red text

## How Has This Been Tested?

Interactively only, checking the error messages

Here only

## Pre-merge checklist

### DIFF
--- a/hutch_python/startup_script.py
+++ b/hutch_python/startup_script.py
@@ -14,6 +14,8 @@ imported from.
 
 def _configure_elog_poster():
     import IPython
+
+    from elog import HutchELog
     from nabs.callbacks import ELogPoster
 
     from hutch_python.utils import safe_load
@@ -22,7 +24,8 @@ def _configure_elog_poster():
         # RE, ELog will already exist by now
         elog = globals().get('elog', None)
         RE = globals().get('RE', None)
-        assert elog is not None
+        if not isinstance(elog, HutchELog):
+            raise RuntimeError("elog not loaded, skip ELogPoster")
 
         elogc = ELogPoster(elog, IPython.get_ipython())
         RE.subscribe(elogc)
--- a/hutch_python/startup_script.py
+++ b/hutch_python/startup_script.py
@@ -14,18 +14,21 @@ imported from.
 
 def _configure_elog_poster():
     import IPython
-
+    from bluesky.run_engine import RunEngine
     from elog import HutchELog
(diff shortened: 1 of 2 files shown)


tangkong:
Seems like elog was never an explicit dependency on the pip-side, weird

ZLLentz:
It must be some issue with a dependency of a dependency that we couldn't get onto pypi

ZLLentz:
For this PR, I'll just protect the import in the safe load or in a try/except

ZLLentz:
The actual answer is probably that `elog` is already taken on pypi

tangkong:
ah yes.  PyPi's elog is an unmaintained elasticsearch log system https://pypi.org/project/elog/

Proposals for names?  `lcls-elog`?  `explog`?  `the-elog`?

hutch_python/startup_script.py
```diff
@@ -19,10 +19,27 @@ def _configure_elog_poster():
...
+            elog = globals()['elog']
+        except KeyError:
+            raise RuntimeError("elog not loaded, skip ElogPoster")
+        try:
+            RE = globals()['RE']
+        except KeyError:
+            raise RuntimeError("RE not loaded, skip ElogPoster")
+        # Even if they exist, things can go wrong if the user accidentally clobbers the name
+        if not isinstance(elog, HutchELog):
+            raise RuntimeError("elog replaced, skip ELogPoster")
```
tangkong, line 40: I guess this is something the hint banner could also fall prey to, if someone overwrites an existing namespace global.  

I think it's maybe a small miracle that we haven't hit upon a case where a hutch overwrites one of our helpful namespaces (`a`, `m`, etc).  Or maybe they just don't use those and access the motors by name explicitly.  🤷

tangkong (approved):
Seems like a good fix.

Issue #334, opened 2022-04-28
Load Additional "Beamlines" into hutch python session


slactjohnson:
## Expected Behavior
It would be nice to be able to load devices from Happi with a bit more granularity than we currently have. 

## Current Behavior
Currently hutches pour every device that they would ever want to use either into the happi database under the same "beamline" location. This is sufficient for now, but I can see issues cropping up once hutches have more than one interaction point (e.g. IP1 and IP2).

## Possible Solution
Add an "endstation" flag or similar to happi and/or the hutch python CLI. This way, hutches could configure beamline devices shared among endstations as "beamline" devices, and devices specific to endstations as "endstation" devices. 

There could be some utility to accepting more than one of these "endstation" or "area" flags (see context below).

I'm open to any other ideas/solutions that would achieve something similar. 

## Context
I'm trying to provide a way to load the MODS devices into laspython. Currently, each MODS is setup as a separate "beamline" in happi to separate the devices in LUCID screens, e.g. `beamline=IP1_MODS`, `beamline=CRIX_MODS`. However, there are also some devices in the `las` "beamline" that they would want in their python sessions, such as beam transport related devices. In the future I can also see wanting to add devices associated with whatever specific laser source is allocated to the hutch as well. It's trivial to create a `laspython` session that loads a specific MODS environment, but it would be nice to have a clean way to do this _and_ load the common `las` devices as well, which I do not see a way to do currently.

## Your Environment
laspython

ZLLentz:
This is a very good idea.

For hutch-python in isolation it'd probably be enough to add config options for loading multiple beamlines. We may even want to take it a step further and allow full control of the happi designations. Would we need to have different variants for one instrument? Using rix an example, do we need to have a section in the config that looks like:
```
happi_variants:
    - ip1:
       - queries:
         - beamline=IP1_MODS
         - beamline=RIX
       - use_lightpath: False
    - crix:
       - queries:
         - beamline=CRIX_MODS
         - beamline=RIX
       - use_lightpath: False
```
And then we do some magic to allow `rix3 ip1` and `rix3 crix` to access the different variants?
Or is enough to do something like
```
happi_queries:
  - beamline=RIX
  - beamline=IP1_MODS
  - beamline=CRIX_MODS
```

One neat benefit to these is that beamlines can opt out of loading upstream devices in their session. You could even have a `bare` variant that skips all `happi` loading.

[...]

PR #395, opened 2025-01-22
ECS-7048 Load Additional "Beamlines" into hutch python session

janeliu-slac:
Added a new feature to allow hutches to load additional devices in hutch-python using conf.yml.

## Description
Currently hutches can only load devices in their own endstation and upstream of their beamline. This feature allows hutches to load additional devices not on their beamline.

## Motivation and Context
https://github.com/pcdshub/hutch-python/issues/334

Currently hutches pour every device that they would ever want to use either into the happi database under the same "beamline" location. Issues come up when hutches have more than one interaction point (e.g. IP1 and IP2) and need to see devices on another beamline.

## How Has This Been Tested?
Manual testing and unit testing

Documentation is in the code and in pre-release notes.

## Pre-merge checklist


hutch_python/happi.py
```diff
@@ -30,7 +30,8 @@ def get_happi_objs(
     light_ctrl: LightController,
     endstation: str,
     load_level: DeviceLoadLevel = DeviceLoadLevel.STANDARD,
-    exclude_devices: list[str] = None
+    exclude_devices: list[str] = None,
+    additional_devices: list = None
```
tangkong, line 34:
```suggestion
    exclude_devices: Optional[list[str]] = None,
    additional_devices: Optional[list] = None
```
This wasn't properly hinted before but while we're here we can be more accurate

hutch_python/happi.py
```diff
@@ -53,16 +54,37 @@ def get_happi_objs(
...
         Name of hutch
 
+    load_level: ``DeviceLoadLevel``
+        load all or standard devices
+
+    exclude_devices: ``list[str]``
+        list of devices that should be excluded when loading
+
+    additional_devices: ``list``
+        devices are loaded based a list containing dictionaries of happi search terms
```
tangkong, line 64: I think it's probably worth being explicit about what the list contains here, since "happi search terms" isn't a specific type.  From a brief lookahead this looks like it should be a `List[Dict[str, str]]`?

tangkong (commented):
A premature review, but I wanted to get eyes on this before I go on vacation

janeliu-slac:
How do you do a happi search in the python code using globs?
```
client = happi.Client.from_config()
results = client.search(name="tmo_*")  # does not work
```
I've only been able to do literal searches, like `client.search(beamline='LAS')`

hutch_python/happi.py
```diff
@@ -53,16 +54,37 @@ def get_happi_objs(
...
         exclude_devices = []
 
+    if not additional_devices:
+        additional_devices = []
+    else:
+        # additional_devices converts the value field of each key-value pair
+        # into a list of strings for easier handling
+        for device_dict in additional_devices.copy():
+            for key, val in device_dict.copy().items():
+                new_val = val.split(',')
```
ZLLentz, line 84: I'm having a bit of deja vu here https://github.com/pcdshub/hutch-python/pull/392#discussion_r1826380485

hutch_python/happi.py
```diff
@@ -118,6 +140,17 @@ def get_happi_objs(
...
             containers.remove(device)
 
+    # Load additional devices
+    for device_dict in additional_devices:
+        for key, value_list in device_dict.items():
+            for value in value_list:
+                reqs = {key: value}
+                results = client.search(**reqs)
+                add_devices = [
+                    res.item for res in results if res.item.name not in dev_names]
```
ZLLentz, line 150: `dev_names` is not guaranteed to be bound here, it's only created in one of the `else` clauses above.

hutch_python/happi.py
```diff
@@ -118,6 +140,17 @@ def get_happi_objs(
         if device.name in exclude_devices:
             containers.remove(device)
 
+    # Load additional devices
+    for device_dict in additional_devices:
+        for key, value_list in device_dict.items():
+            for value in value_list:
+                reqs = {key: value}
+                results = client.search(**reqs)
```
ZLLentz, line 148: This is a big confusing to me. It looks like we're running each search separately and including all the devices from each of them, but this doesn't allow anyone to do a more specific search, e.g. for devices that are both `name=tmo.*` and `beamline=las` at the same time. I think there needs to be a way for someone to run multiple different specific searches.

tangkong: We discussed this a bit during the dev sync, but I forgot to mention it here.  Zach is right, there should be a way to search for the **intersection** of a group of search terms, not just the **union** of these search terms

janeliu-slac: Yep, I got stuck on this part earlier but working on a solution to allow for searches with globs in happi.py (and allow for unions and intersections of search terms in the results).

hutch_python/happi.py
janeliu-slac: I revised conf.yml to use happi search criteria (globs) and get union/intersection results. In happi.py `additional_devices` now has the type `Dict[str, Union[str, bool]]`.

It looks like using the `typing` module might be generally preferable to static type checking. Python's static type checking can silently fail in a few situations, but the `typing` module always reports an error.

tangkong: No numeric values allowed? 

Typing in python is entirely optional, and all types are type-hints. So to be totally correct here, python doesn't have native static type checking.  You can use an external library like `mypy`, but python itself doesn't check any of the type hints you provide. 

What exactly is throwing an error here?  Generally if you are using capitalized type aliases (`List`, `Dict` , `Set`, etc) you have to import them from typing

janeliu-slac:
I finally found the error in my unit test set up. In test_happi.py both `exclude_devices` and `additional_devices` parameters are passed into `get_happi_objs()` but I didn't specify which type of parameter it was and just passed in the variable name. I added `exclude_devices=exclude_devices` and  `additional_devices=additional_devices` to `get_happi_objs()` depending on the unit test type and that fixed it.

hutch_python/happi.py
janeliu-slac: Updated additional_devices in load_conf.py to be handled when it's first received.

hutch_python/happi.py
janeliu-slac: Updated the type for `additional_devices` and got it to accept numerical values like `z = -1, 1`.

hutch_python/happi.py
```diff
@@ -118,6 +135,28 @@ def get_happi_objs(
         if device.name in exclude_devices:
             containers.remove(device)
 
+    # Load additional devices
```
ZLLentz, line 138: Maybe exclude should happen after additional so we can exclude from additional
This gets a little confusing but I think we should change the order

hutch_python/happi.py
```diff
@@ -118,6 +135,28 @@ def get_happi_objs(
...
+        for entry_key, entry_val in search_val.items():
+            search_string = entry_key + '=' + entry_val
+            search_list.append(search_string)
+
+        search_results = search_parser(
+            client=client,
+            use_glob=True,
+            search_criteria=search_list
+        )
+        search_results = list(search_results)
```
ZLLentz, line 153: Reviewer note: this is correct, the annotations on `search_parser` are wrong! Sometimes `search_parser` gives us a `set`, even though it claims to always give a `list`

ZLLentz (dismissed):
Code is good, just that one question/quibble about addition/exclude order. Good job!

hutch_python/happi.py
janeliu-slac: I'll add this to my notes for follow up (to change the annotations).

ZLLentz (approved):

PR #396, opened 2025-02-03
ECS-7048 In yaml_files.rst fixed a typo

janeliu-slac:
## Description
I couldn't git commit to master directly (don't have permission), so created this branch to fix the typo in yaml_files.rst.


ZLLentz (approved):
LGTM

No need to add e.g. a pre-release notes since this will be part of the same tag as the previous

PR #397, opened 2025-03-14
BLD: update tests/code to be compatible with new ipython >= 8.26.0

tangkong:
## Description
A couple of quick patches to make hutch-python compatible with newer IPython versions.
- Updates IPythonSessionTimer cell hooks to accept any and all arguments.  At some point they changed this to actually take information payloads, but we don't use them and I don't want it to break again
- Uses the IPCompleter in our tab completion tests, and spoofs the line buffer.  This seems to actually be what is in the buffer based on my debugging.  It's eventually supposed to be deprecated itself, but this is 

## Motivation and Context
Newer IPython versions are needed to move to python 3.12.

IPython is good about communicating its deprecated methods, but not the best at actually describing its internal api.  I suppose we're not really supposed to be mucking around with its autocompleters manually anyway.

I think if we do decide to pin the environment wide ipython version, it's reasonable to do so in pcds-envs.  Ipython incompatibility (due to scientist protest at least) isn't a fundamental incompatibility with hutch-python, so it might make sense to not pin it here.

## How Has This Been Tested?
Interactively, and by making sure the test suite runs.

This PR

## Pre-merge checklist


tangkong:
Before I question my entire existence, I wanted to ago and verify what the minimum versions for each ipython/python were.
According to the git blame on conda-feedstock/ipython-feedstock
* py3.11  -> as of IPython 9.0.0
* py3.10 -> as of IPython 8.19.0
  * (digging further back, since conda meta variables were only set in 8.30.0)

tangkong:
I think placing upper pin at ipython 9 is reasonable.  There's not much functional in ipython 9, aside from llm autosuggestion support through jupyterlab.

https://ipython.readthedocs.io/en/stable/whatsnew/version9.html

I think this pin should be sufficient to allow us to deal with our other dependencies for a while (line-profiler? I forget)

hutch_python/tests/test_cli.py
```diff
@@ -160,7 +160,8 @@ def __dir__(self):
     # us.
     hutch_python.cli.configure_ipython_session(HutchPythonArgs())
 
-    completer = IPython.core.completer.Completer(namespace=ns)
+    completer = IPython.core.completer.IPCompleter(namespace=ns)
```
ZLLentz, line 163: Is there a secret minimum pin lurking in this rename?

ZLLentz (approved):

hutch_python/tests/test_cli.py
ZLLentz: If there is I suppose it'd be hard to find

hutch_python/ipython_session_timer.py
```diff
@@ -58,11 +58,11 @@ def __init__(self, ipython):
         ipython.events.register('pre_run_cell', self._set_user_active)
         ipython.events.register('post_run_cell', self._set_user_inactive)
 
-    def _set_user_active(self):
+    def _set_user_active(self, *args, **kwargs):
```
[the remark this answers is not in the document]
janeliu-slac: Nevermind, I googled my own question :)

_In IPython 9.0, when registering a callback function using ipython.events.register, the callback function must accept arguments, even if those arguments are not used within the function's body. The specific arguments required depend on the event for which the callback is registered.

janeliu-slac (approved):

hutch_python/tests/test_cli.py
tangkong: Git blame for the `IPCompleter` dates back 18 years, to before the git history started
image

PR #399, opened 2025-04-15
MNT: Adding CODEOWNERS file

tangkong:
(autogenerated) adding codeowners file


tangkong:
sigh conda tests are timing out, I'll look at that in a later effort

PR #403, opened 2026-02-10
MNT: replace deprecated pkg_resources with importlib.metadata.distributions

tangkong:
## Description
- pkg_resources was fully deprecated in setuptools 82, and builds are starting to pick this up
- importlib.metadata has existed since py3.8 apparently

## Motivation and Context
pcds-envs had hutch-python fail here [recently](https://github.com/pcdshub/pcds-envs/actions/runs/21812496319/job/63034597274#step:6:3784).

## How Has This Been Tested?
pytest, and opening hutch-python in --debug mode to see that there's still sensible env dump information

## Where Has This Been Documented?
This PR

## Pre-merge checklist
- [x] Code works interactively (a real hutch config file can be loaded)
- [x] Code contains descriptive docstrings, including context and API
- [x] New/changed functions and methods are covered in the test suite where possible
- [x] Test suite passes locally
- [x] Test suite passes on GitHub Actions
- [x] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [x] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #404, opened 2026-03-27
MNT: re-enable jedi completion for environments that use newer IPython builds

ZLLentz:
## Description

See title

## Motivation and Context

The default completer for IPython is Jedi, but we had disabled it because it inspects the types of properties by evaluating the code, which can cause large slowdowns. In newer version of IPython, this is no longer the case: property evaluation is not done as part of tab completion by default (you can turn it on).

In our deployed version of IPython, we're seeing some weird behavior for autocompletion, which is not present if we re-enable jedi.

https://jira.slac.stanford.edu/browse/ECS-10047

## How Has This Been Tested?

Interactively only

I've ended up disabling/skipping one test in the suite, for a few reasons:

1. I couldn't figure out how to fix it within an hour- I managed to get a more realistic completer going but it was broken for other reasons I don't want to keep trying
2. The IPython completer API is really unstable- many functions are labelled deprecated or experimental
3. This test didn't even catch our issue last time

It will only be in the release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/404-mnt_enable_jedi.rst
+++ b/docs/source/upcoming_release_notes/404-mnt_enable_jedi.rst
@@ -1,0 +1,23 @@
+404 mnt_enable_jedi
+###################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Work around an issue where non-jedi completions were broken for pcdsdevices
+  in certain IPython ranges by re-enabling jedi completions in those ranges.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zllentz
--- a/hutch_python/cli.py
+++ b/hutch_python/cli.py
@@ -85,6 +85,14 @@ def configure_tab_completion(ipy_config):
     """
     Disable Jedi and tweak IPython tab completion.
 
+    At some IPython version this became no longer needed due to fixed performance issues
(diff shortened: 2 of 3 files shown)


tangkong:
Any time we mess with ipython autocompletion I am filled with fear.  It does look like jedi's been subject to quite a few changes over the last few ipython versions.

I'll take this for a ride in lieu of digging through source/release notes.

tangkong:
Downgrading to a pcds-conda that uses old ipython doesn't work to test, since there were some breaking changes with distutils https://github.com/pcdshub/hutch-python/pull/403.

I tried running current master on pcds-6.1.0 and autocompletion still seemed to work for me.  Maybe I'm not reproducing this correctly?...

Other than that the code looks good to me.

ZLLentz:
> I tried running current master on pcds-6.1.0 and autocompletion still seemed to work for me

Specifically, `ipython` with `use_jedi = False` fails at nested tab complete as described in https://jira.slac.stanford.edu/browse/ECS-10047

tangkong:
I tried this, but maybe I did it wrong 🤔 

Indeed I did, I didn't realize it was only an issue with **nested** tab complete.

tangkong (approved):
Looks good to me in my testing as well

ZLLentz:
I think the important thing is that the jedi performance in ipython is much better now with our objects due to some of the other changes that they made, and using the mode that is clearly tested more often is preferrable for us

